PR #1, opened 2022-08-13
ItemStackMap

mitchej123:
Should work for Hodgepodge Furnace Fix (NBT insensitive option), and any other place that needs to use it.


mitchej123:
Still need to test, but the design seems solid.  Plan to use this to remove the GT5u dep on hodgepodge, and the need to copy paste it anywhere else

Connor-Colenso (approved):
LGTM!

PR #2, opened 2022-08-15
Make it load right after mixins

vlaetansky:


Glease:
I don't think this will ever work... @ Mod aren't even parsed when mixins are run

draknyte1:
> I don't think this will ever work... @ Mod aren't even parsed when mixins are run

Do you need it to be? Could just specify after:forge and before:* (if wildcard doesn’t work, then just make sure it’s before the other mods or have them after:this.

Glease:
In the log dream sent me the normal mod parsing hasn't even started yet and there are already a bunch of ClassNotFoundException from mixin. Only thing completed is coremod discover.

Glease:
Mixin finish the transform before normal mod loading even begins. It then cache the transformation result and reuse it later on whenever the class being transformed would be loaded. I know mitch has a trick to load jar early but I don't know how. I suggested to let him have a look but apparently he is super busy now...

mitchej123:
I think we need to make it an empty coremod so that it's available early on.  That's (my guess) as to why hodgepodge previously worked injecting this. I'm in meetings all morning, but will look at it this evening if nobody else has had a chance.

mitchej123:
```
[17:43:11] [main/WARN] [mixin]: Error loading class: com/gtnewhorizon/gtnhlib/util/map/ItemStackMap (java.lang.ClassNotFoundException: The specified class 'com.gtnewhorizon.gtnhlib.util.map.ItemStackMap' was not found)
[17:43:11] [main/INFO] [STDERR]: [java.lang.Throwable$WrappedPrintStream:println:-1]: java.lang.RuntimeException: java.lang.ClassNotFoundException: com.gtnewhorizon.gtnhlib.util.map.ItemStackMap
[17:43:11] [main/INFO] [STDERR]: [java.lang.Throwable$WrappedPrintStream:println:-1]:     at org.spongepowered.asm.mixin.transformer.MixinPreProcessorStandard.transformMethod(MixinPreProcessorStandard.java:682)
```

The relevant piece of the log in my mind.  The GTNHLib hasn't loaded yet, but hodgepodge injected the class call.  Either a coremod to load, or hodgepodge needs to force load it.

vlaetansky:
Not sure what you mean by an empty coremod, i have look at hodgepodge and decided that this is what you meant to do.

mitchej123:
Examples: https://github.com/GTNewHorizons/Hodgepodge/blob/master/src/main/java/com/mitchej123/hodgepodge/asm/HodgePodgeASMLoader.java
https://github.com/GTNewHorizons/NewHorizonsCoreMod/blob/master/src/main/java/com/dreammaster/coremod/DreamCoreMod.java


Anything that'll trigger the coremod loading from FML

PR #3, opened 2022-08-27
Implement some useful optional features to make this faster

Glease:
Implementation defined in superinterface's default method does not know about the internals of map, thus operate with less efficiency, e.g. merge() must do 2 hash lookup instead of one. Implementing these in subclass allow better performance should client code take use of these optional features.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/util/map/ItemStackMap.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/util/map/ItemStackMap.java
@@ -552,10 +552,10 @@ public final class ItemStackMap<T> extends AbstractMap<ItemStack, T> {
     }
 
     private enum KeyType {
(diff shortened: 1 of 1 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/util/map/ItemStackMap.java
```diff
@@ -461,6 +479,54 @@ private void updateSize() {
                 size = newSize;
             }
         }
+
+        public T merge(ItemStack key, T value, BiFunction<? super T, ? super T, ? extends T> remappingFunction) {
+            try {
+                switch (getKeyType(actualDamage(key), key.stackTagCompound)) {
+                    case 0:
```
eigenraven, line 486: I know it already was like this, but can you make the key types an enum?

eigenraven (approved):

PR #6, opened 2022-09-04
Add debug messages methods

Alexdoru:


eigenraven (changes requested):
The methods have to be defined in CommonProxy too, and ClientProxy should `@Override` them. They have to be callable from the GTNHLib.proxy object directly, otherwise you force-load ClientProxy if you try to cast to it which can crash the server

Alexdoru:
done

eigenraven (approved):

PR #11, opened 2022-10-21
Add API for server to tell client to show message above hotbar

miozune:


boubou19:
what's the difference with #7 ?

miozune:
`printMessageAboveHotbar` works only with client. If server wants client to show message (e.g. block activation), server needs to send packet to tell client to do that. This PR contains the latter.

src/main/java/com/gtnewhorizon/gtnhlib/ClientProxy.java
```diff
@@ -112,4 +115,17 @@ public void addMessageToChat(IChatComponent componentText) {
...
     }
+
+    @Override
+    public void sendMessageAboveHotbar(
+            EntityPlayerMP player,
+            IChatComponent chatComponent,
+            int displayDuration,
+            boolean drawShadow,
+            boolean shouldFade) {
+        if (FMLCommonHandler.instance().getEffectiveSide() == Side.CLIENT) {
```
Glease, line 126: I can't think of a situation where you have a EntityPlayerMP but is on logical client.

src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
```diff
@@ -49,4 +54,18 @@ public void addErrorToChat(IChatComponent componentText) {}
...
+     * Sends packet from server to client that will display message above hotbar.
+     * @see ClientProxy#printMessageAboveHotbar
+     */
+    public void sendMessageAboveHotbar(
+            EntityPlayerMP player,
+            IChatComponent chatComponent,
+            int displayDuration,
+            boolean drawShadow,
+            boolean shouldFade) {
+        NetworkHandler.instance.sendTo(
```
Glease, line 68: As a precaution, please rule out fake player here.

draknyte1: Also remember TC doesn’t extend FakePlayer.

src/main/java/com/gtnewhorizon/gtnhlib/ClientProxy.java
miozune: Indeed, `EntityPlayerMP` requires `MinecraftServer` and `WorldServer` for instantiation

src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
miozune: sure

draknyte1: @miozune still doesn’t cover TC fake players.

miozune: Now I think, is this check really needed? For example GG does send message without FakePlayer check:
https://github.com/GTNewHorizons/GoodGenerator/blob/bd60bd3e96368d1fe726cf501ce1166b397d896a/src/main/java/goodgenerator/blocks/tileEntity/GTMetaTileEntity/NeutronSensor.java#L102

eigenraven (approved):
The check is useful for defensive coding to prevent crashes by api misuse, but not strictly necessary. We should just fix TC via mixins to make its fake player derive the FakePlayer class, not add workarounds in other mods

PR #12, opened 2022-12-18
Add Tooltip Animation API

glowredman:
Extracted from https://github.com/GTNewHorizons/GT5-Unofficial/blob/master/src/main/java/gregtech/client/GT_TooltipEventHandler.java.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/util/AnimatedTooltipHandler.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/util/AnimatedTooltipHandler.java
@@ -1,0 +1,185 @@
+package com.gtnewhorizon.gtnhlib.util;
+
+import cpw.mods.fml.common.eventhandler.SubscribeEvent;
+import cpw.mods.fml.common.registry.GameRegistry;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+import net.minecraft.item.Item;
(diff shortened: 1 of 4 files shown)


Glease (commented):
How about using a ItemStackMap for tooltipMap instead?

src/main/java/com/gtnewhorizon/gtnhlib/util/AnimatedTooltipHandler.java
```diff
@@ -0,0 +1,184 @@
...
+     * @param posstep How many steps {@code formattingArray} is shifted each {@code delay}
+     * @param delay How many milliseconds are between each shift of {@code formattingArray}
+     * @param formattingArray An array of formatting codes. Each char of {@code text} will be prefixed by one entry, depending on {@code posstep} and {@code delay}. Wraps around, if shorter than {@code formattingArray}.
+     * @author TTFTCUTS, glowredman
+     */
+    public static Supplier<String> animatedText(String text, int posstep, int delay, String... formattingArray) {
+        if (text == null || text.isEmpty() || formattingArray == null || formattingArray.length == 0) return () -> "";
+
+        final int finalDelay = delay > 0 ? delay : 1;
+        final int finalPosstep = posstep >= 0 ? posstep : 0;
```
OneEyeMaker, line 81: I suggest to use `Math.max` for readability.

OneEyeMaker (approved):

PR #17, opened 2023-01-22
Add (final static and others) field reflection utilities, compatible with Java 8-19

eigenraven:
A few mods use hacky code to change the `modifiers` field of `java.lang.reflect.Field` instances, this doesn't work as of Java 12. As the replacement is not trivial, I think putting this in shared code makes more sense, and I can provide a nicer optionally type-checked API for it too.

Includes some simple JUnit tests, I have not tested every primitive type because copy-pasting the same tests 9 times seems a bit silly, but I can't think of another solution. I ran the tests on java versions 8, 11, 17 and 19 and all of them work.

It uses Unsafe, but afaik in modern Java there is no other alternative to write to `final static` fields without removing `final` via class transformation before loading.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/reflect/Fields.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/reflect/Fields.java
@@ -1,0 +1,391 @@
+package com.gtnewhorizon.gtnhlib.reflect;
+
+import com.google.common.base.Throwables;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.reflect.Modifier;
+import java.util.Objects;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import sun.misc.Unsafe;
+
+/**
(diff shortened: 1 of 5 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/reflect/Fields.java
```diff
@@ -0,0 +1,391 @@
...
+                try {
+                    // Generate getter
+                    final MethodHandle getterHandle = mhLookup.unreflectGetter(this.javaField);
+                    if (this.isStatic) {
+                        final MethodHandle exactGetterHandle = getterHandle.asType(MethodType.methodType(Object.class));
+                        this.getAccessor = obj -> {
+                            try {
+                                return (F) exactGetterHandle.invokeExact();
+                            } catch (Throwable e) {
+                                Throwables.propagate(e);
```
miozune, line 272: `throw Throwables.propagate(e);`, c.f. https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate

eigenraven: Good catch, I learned about this from forge code and it seems they are not throwing it in many places where they should :P

miozune (approved):

PR #23, opened 2024-01-12
Port configs from Angelica/ArchaicFix

Caedis:
This will allow our mods to easily create config files as well as allow them to be editable in game from the Forge config menu.

Should we also move this to here or even HP?
https://github.com/GTNewHorizons/EnderCore/blob/master/src/main/java/com/enderio/core/client/gui/bettermodlist/EventHandlerGui.java

This allows the forge config gui to work in game.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,9 @@ GTNHLib
 ## About
 
 Shared code library for GTNH mods
+
+
+
+## Credits
+
+Configs: Originally by FalsePattern, licensed under LGPL-3.0 in [FalsePatternLib](https://github.com/FalsePattern/FalsePatternLib/tree/master/src/main/java/com/falsepattern/lib/config)
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -9,6 +9,10 @@ dependencies {
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine")
     testRuntimeOnly("org.junit.vintage:junit-vintage-engine")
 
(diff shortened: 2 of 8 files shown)


Glease:
1. lombok? if possible, at least make sure it doesn't got passed down  the dependency chain
2. where does this piece of code come from? I'd imagine it's not yours code, nor code of anyone inside GTNH org. proper credit need to be given somehow

Caedis:
It all came from https://github.com/GTNewHorizons/Angelica/tree/master/src/main/java/com/gtnewhorizons/angelica/config so ask @mitchej123

Caedis:
How would I make sure deps dont add it?

Caedis:
Looks like the original source is AF, which is LGPL-3
https://github.com/embeddedt/ArchaicFix/tree/main/src/main/java/org/embeddedt/archaicfix/config

bombcar:
> How would I make sure deps dont add it?

According to https://wiki.gtnewhorizons.com/wiki/Development#Dependency_types you have the right one, it shouldn't be pulled into further things downstream.

mitchej123:
Should probably add something to Credits indicating where it came from and the license.  I didn't realize it came to AF via FPL or I would have included that in Angelica's credits.

mitchej123:
Yah, talked to FP - he's fine with it going here, but let's put a mention in the credits/readme giving them credit

Caedis:
I actually had that ready to commit/push for awhile

mitchej123 (approved):

PR #24, opened 2024-01-13
Add fastutil to dependencies.gradle

tth05:
Already used in Angelica, although it can be minimized there. This way it can be re-used everywhere though.
Bloats the file size from 50kb to 22MB 💀


mitchej123:
Consider this tentative approval -- still collecting some thoughts from other people, and would like @eigenraven to chime in

eigenraven (approved):

PR #26, opened 2024-01-29
ModDetectedDefault config annotation

Caedis:


Caedis:
Tested with MouseTweaks with GT5U loaded. Default for WheelTweak is true, but tested if gregtech is loaded, it was false as expected.

Caedis:
TODO:  

- [ ] Cache the output for classes and mod load status

src/main/java/com/gtnewhorizon/gtnhlib/config/ConfigurationManager.java
```diff
@@ -92,36 +91,128 @@ private static void processConfigInternal(Class<?> configClass, String category,
...
+                }
                 field.setInt(null, rawConfig.getInt(name, category, defaultValue, min, max, comment, langKey));
             } else if ((boxed = fieldClass.equals(Float.class)) || fieldClass.equals(float.class)) {
                 val range = Optional.ofNullable(field.getAnnotation(Config.RangeFloat.class));
                 val min = range.map(Config.RangeFloat::min).orElse(Float.MIN_VALUE);
                 val max = range.map(Config.RangeFloat::max).orElse(Float.MAX_VALUE);
-                val defaultValue = Optional.ofNullable(field.getAnnotation(Config.DefaultFloat.class))
+                float defaultValue = Optional.ofNullable(field.getAnnotation(Config.DefaultFloat.class))
                         .map(Config.DefaultFloat::value).orElse(boxed ? (Float) field.get(null) : field.getFloat(null));
+                if (field.isAnnotationPresent(Config.ModDetectedDefault.class)) {
```
mitchej123, line 147: This seems repeated amongst all of the cases - would be nice to turn it into a function/class helper that could be used in the chain above...

src/main/java/com/gtnewhorizon/gtnhlib/config/ConfigurationManager.java
```diff
@@ -92,36 +91,128 @@ private static void processConfigInternal(Class<?> configClass, String category,
...
                 field.setFloat(null, rawConfig.getFloat(name, category, defaultValue, min, max, comment, langKey));
             } else if (fieldClass.equals(String.class)) {
-                val defaultValue = Optional.ofNullable(field.getAnnotation(Config.DefaultString.class))
+                String defaultValue = Optional.ofNullable(field.getAnnotation(Config.DefaultString.class))
                         .map(Config.DefaultString::value).orElse((String) field.get(null));
+                if (field.isAnnotationPresent(Config.ModDetectedDefault.class)) {
+                    val newDefault = Optional.ofNullable(field.getAnnotation(Config.ModDetectedDefault.class))
+                            .map(f -> {
+                                if (!f.coremod().isEmpty()) {
+                                    try {
```
mitchej123, line 173: I'd think caching these in a fastutils <something>2Boolean would be a good idea

Caedis:
Closing for now, will probably be rewritten if its still needed

PR #36, opened 2024-02-28
Add a compare-and-swap reference wrapper and collections

eigenraven:
Useful for situations of very rare mutations and very frequent concurrent reads.
The CAS wrapper can be used to easily implement it for any fastutil collection, the generic CasList and CasMap serve as a utility for Object collections and implementation reference.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/concurrent/cas/CasList.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/concurrent/cas/CasList.java
@@ -1,0 +1,328 @@
+package com.gtnewhorizon.gtnhlib.concurrent.cas;
+
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Spliterator;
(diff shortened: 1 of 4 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/concurrent/cas/CasList.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    /**
+     * @deprecated Basically useless as it could be modified after being checked, iterate over a stored reference from
+     *             {@link CasList#read()} instead.
+     */
+    @Override
+    @Deprecated
+    public int size() {
+        return read().size();
```
embeddedt, line 78: Should these even be deprecated or should they just throw UnsupportedOperationException?

eigenraven: I'd like for this to work as a drop-in, and theoretically it's fine when you use size() only to report the current size or do some other metrics that aren't just iterating in a for-loop

mitchej123 (approved):

PR #38, opened 2024-04-11
Add suffixes 's' for stacks and 'i' for ingots of fluid to MathExpressionParser.

AbdielKavash:
Similarly to existing 'k', 'm', etc. suffixes, one can now type "4s" for 256 (4 stacks of typical items), or "8i" for 1,152 (8 ingots' worth of fluids, 8\*144).

This should work everywhere that uses the parser; i.e., ModularUI and AE2 terminals.


Caedis:
Have it be S for Stacks

I could see `s` being added in the future for as a shorthand for seconds where ticks can be inputted

AbdielKavash:
Right now both 's' and 'S', and both 'i' and 'I' work.

All suffixes right now are case-insensitive (though it is possible to make different cases do different things in the future, if desired).

Caedis:
Guess that works too

Caedis (approved):

PR #47, opened 2024-08-10
Vertex Normals for the capturing tesselator

CookieBrigade:
With this the capturing tesselator can be used to upload vertex normals to render out.

I'm not sure if the captureVertexNormals should be settable in the tesselator or the Wavemodel Mixim, but the former seemed easier to leak the flag after use.  If there's a solution without a threadlocal flag I'd like to see how I'd avoid them.


CookieBrigade:
I'm going to close this because I think I have a better solution, but I need some time to cook with it.

PR #49, opened 2024-08-13
Backport BlockPos#getAllInBox

miozune:


eigenraven (changes requested):
Please add JavaDoc comments explaining that the bounds are inclusive (`[x1,x2]` and not for example `[x1,x2)`)

Caedis:
Might I suggest adding a test if this repo is setup for them?

eigenraven (approved):
Looks good, I wonder if we should also add a Stream factory to match the iterator version with the "new" api (Spliterators are also a bit more efficient than classic Iterators)

PR #56, opened 2024-08-22
Fix transformer exclusions when loading without RFB

FalsePattern:
They also need to be checked when RFB is not present and the transformer is going through the legacy transformer pipeline.


mitchej123 (approved):

Caedis:
Can you run spotless on it?

PR #58, opened 2024-08-31
Backport 1.12.2 config subcategory behavior

Lyfts:
In 1.12.2 you can declare static instances of inner classes in your `@Config` annotated class, those instances will then be automatically treated and parsed as subcategories of that class.

There's an example of it here https://github.com/GTNewHorizons/GTNHLib/blob/056959f0f51e3f2699274d264a9a7887a257f315/src/main/java/com/gtnewhorizon/gtnhlib/config/TestConfig.java

This will automagically be parsed into

```
# Configuration file

test {
    # This is a test boolean [default: true]
    B:testBoolean=true

    # This is a test double [range: 4.9E-324 ~ 1.7976931348623157E308, default: 1.0]
    D:testDouble=1.0

    # This is a test enum
    # Possible values: [TEST1, TEST2, TEST3]
    #  [default: TEST1]
    S:testEnum=TEST1

    # This is a test int [range: -2147483648 ~ 2147483647, default: 1]
    I:testInt=1

    ##########################################################################################################
    # subconfig
    #--------------------------------------------------------------------------------------------------------#
    # This is a sub config
    ##########################################################################################################

    subconfig {
        # This is a test boolean [default: true]
        B:testBoolean=true

        # This is a test double [range: 4.9E-324 ~ 1.7976931348623157E308, default: 1.0]
        D:testDouble=1.0

        # This is a test enum
        # Possible values: [TEST1, TEST2, TEST3]
        #  [default: TEST1]
        S:testEnum=TEST1

        # This is a test int [range: -2147483648 ~ 2147483647, default: 1]
        I:testInt=1

        ##########################################################################################################
        # doublesubconfig
        #--------------------------------------------------------------------------------------------------------#
        # This is a nested sub config
        ##########################################################################################################

        doublesubconfig {
            # This is a test boolean [default: true]
            B:testBoolean=true

            # This is a test double [range: 4.9E-324 ~ 1.7976931348623157E308, default: 1.0]
            D:testDouble=1.0

            # This is a test enum
            # Possible values: [TEST1, TEST2, TEST3]
            #  [default: TEST1]
            S:testEnum=TEST1

            # This is a test int [range: -2147483648 ~ 2147483647, default: 1]
            I:testInt=1
        }

    }

}
```

I also made it possible to get categorized IConfigElements for the config gui so that it will look like this

gtnhlib_conf

As opposed to something like Hodgepodge's current config gui where all the elements are on one page with a humongous scrollbar.

Also adds a method to save the current field values to the config file & adds double as a config option, because why tf is double[] an option while double isn't


Caedis:
Think its time to refactor `processConfigInternal`  

Maybe add a method to the annotations so they the logic is there and we just pass it whatever it needs

Caedis (approved):

PR #73, opened 2024-09-14
Fix a few config issues

Lyfts:
1. Fixes subcats having their category name twice in the langkey & stops subcats from having their lang key set multiple times.
2. Makes `@Config.Comment` applicable to classes so that you can put the annotation on the subcat class instead of their field & add comments for non-subcategories
3. Adds the `-Dgtnhlib.printkeys` jvm flag to print lang keys generated from `@Config.LangKeyPattern` in the log.


boubou19:
tested in https://github.com/GTNewHorizons/GT5-Unofficial/pull/3169 and everything works now thx.

boubou19 (approved):

PR #79, opened 2024-10-05
Auto create config gui if no gui factory is registered

Lyfts:
This automatically creates a config gui with every registered config for a given mod. Since that leaves no options to control which config classes appear in the ui, I did add the `@Config.ExcludeFromAutoGui` annotation which serves as a simple way to keep that control intact.

If no custom config gui is needed then this effectively removes the need to register a gui factory.


Caedis:
How has this been tested?

Caedis (approved):
Good change

src/main/java/com/gtnewhorizon/gtnhlib/mixins/Mixins.java
```diff
@@ -18,6 +18,10 @@ public enum Mixins {
             .setApplyIf(() -> true).addMixinClasses("MixinTessellator")),
     WAVEFRONT_VBO(new Builder("WavefrontObject").addTargetedMod(TargetedMod.VANILLA).setSide(Side.CLIENT)
             .setPhase(Phase.EARLY).setApplyIf(() -> true).addMixinClasses("MixinWavefrontObject")),
+
+    GUI_MOD_LIST(new Builder("Auto config ui").addTargetedMod(TargetedMod.VANILLA).setSide(Side.CLIENT)
+            .setPhase(Phase.EARLY).addMixinClasses("fml.MixinGuiModList"));
```
OneEyeMaker, line 23: There should be comma instead of semicolon at the end of the line.

PR #80, opened 2024-10-07
Backport @EventBusSubscriber annotation from 1.12.2

Lyfts:
In 1.12.2 the `@EventBusSubscriber` annotation can be applied to a class to automatically register it on the MinecraftForge eventbus. In our ancient version we have no less than 4 official eventbusses so instead it will register the methods on the relevant event bus. This works without an instance and thus all methods annotated with `@SubscribeEvent` in the class are expected to be static.

Registering them statically through the standard `EventBus.register` is not ordinarily possible in 1.7.10 so the registration part is done in house. Going about this in the same way that Forge does it (via yucky reflection) proved to have a few issues, it is both slow and will cause the entire method table to be loaded. This in turn loads the class of all the method parameters which is a real problem since you can't put the annotation behind a `Loader.isModLoaded` check, meaning an annotated class that has an event from a mod that isn't loaded will crash the game unless the method is annotated with `@Optional.Method`.

In my eyes requiring any event from a different to be annotated with  `@Optional.Method` is not viable, which meant I had to get a little creative with how the subscriber classes are parsed. Since reflection is out of the game and mixins aren't usable for this, that left ASM as the top option. It uses a coremod transformer (which is less of a transformer and more of a reader) to comb through the methods of the annotated class finding the ones that need to be subscribed to an event. 

This is overall much less disgusting than doing it with reflection and doesn't load anything from the annotated class which in turn allows for any event to be present even if the mod that it's from isn't. The method will never be invoked and invoking the remaining methods in the class won't cause the JVM to try to load that event either.


Lyfts:
I gave it a bit of a torture test by applying it to a large part of the functionally static event handlers in GT5U https://github.com/GTNewHorizons/GT5-Unofficial/tree/eventbussub and everything seems to be working swimmingly so far.

src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
```diff
@@ -64,4 +84,15 @@ public String getMixinConfig() {
     public List<String> getMixins(Set<String> loadedCoreMods) {
         return Mixins.getEarlyMixins(loadedCoreMods);
     }
+
+    @Override
+    public boolean registerBus(EventBus bus, LoadController controller) {
```
mitchej123, line 89: lol, I had no idea this was even a thing

mitchej123: Also - is this something generally needed?  Not immediately clear why it's needed here vs your gt5u example pr

src/main/java/com/gtnewhorizon/gtnhlib/core/transformer/EventBusSubTransformer.java
```diff
@@ -0,0 +1,157 @@
...
+            }
+
+            if (condition) {
+                if (mn.desc.equals("()Z")) {
+                    EventBusUtil.getConditionsToCheck().put(transformedName, mn.name + mn.desc);
+                }
+                continue;
+            }
+
+            if (subscribe == null) continue;
```
mitchej123, line 95: Worth an error log if we get this far?

mitchej123 (approved):
@Dream-Master let @Lyfts merge in case he wants to address any of the comments I left.

mitchej123:
One other thing: Please confirm you've tested on RFB + Non-RFB

src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
Lyfts: Its only purpose is to fetch the classes the transformer needs to visit before the regular mods are constructed, this is needed for the annotation to work in the @Mod annotated class along its proxies.

Lyfts:
It's tested with and without RFB on both server & client

mitchej123:
@Lyfts Good from my end - feel free to either merge, or remove the `don't merge` label

PR #84, opened 2024-10-21
merge iih + mt hand command in one

Alexdoru:
and move it to GTNHlib so that it's available from every dev environement when using client 17

related pr : https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1051

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/commands/ItemInHandCommand.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/commands/ItemInHandCommand.java
@@ -1,0 +1,87 @@
+package com.gtnewhorizon.gtnhlib.commands;
+
+import static net.minecraft.util.EnumChatFormatting.*;
+
+import java.util.Collections;
+import java.util.List;
+
+import net.minecraft.client.entity.EntityClientPlayerMP;
+import net.minecraft.command.ICommandSender;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.fluids.IFluidContainerItem;
+import net.minecraftforge.oredict.OreDictionary;
+
+import cpw.mods.fml.common.registry.GameRegistry;
+
+public class ItemInHandCommand extends GTNHClientCommand {
+
+    @Override
+    public String getCommandName() {
+        return "iteminhand";
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


Alexdoru:
image

src/main/java/com/gtnewhorizon/gtnhlib/commands/ItemInHandCommand.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+        // spotless:on
+    }
+
+    private void printFluidContent(ItemStack itemStack) {
+        if (itemStack.getItem() instanceof IFluidContainerItem tFluidContainer) {
+            FluidStack fluidStack = tFluidContainer.getFluid(itemStack);
+            if (fluidStack != null) {
+                String s = String.format(
+                        "FluidID: [%d], UnlocName: [%s], Name: [%s]",
```
glowredman, line 57: For the item itself, `Unloc.Name` is used, this should be unified (I personally would like to see `Unlocalized Name` but maybe that's to long.

src/main/java/com/gtnewhorizon/gtnhlib/commands/ItemInHandCommand.java
```diff
@@ -0,0 +1,86 @@
...
+        if (iCommandSender instanceof EntityClientPlayerMP player) {
+            ItemStack itemStack = player.getCurrentEquippedItem();
+            if (itemStack == null) {
+                addChatMessage("Hand is empty!");
+                return;
+            }
+            GameRegistry.UniqueIdentifier UID = GameRegistry.findUniqueIdentifierFor(itemStack.getItem());
+            addChatMessage(GREEN.toString() + STRIKETHROUGH + "--------------------" + RESET + " Item info " + GREEN + STRIKETHROUGH + "--------------------");
+            addChatMessage(String.format(GREEN + "Unloc.Name:" + RESET + " [%s]", itemStack.getUnlocalizedName()));
+            if (UID != null) addChatMessage(String.format(GREEN + "ItemName:" + RESET + " [%s]", UID));
```
glowredman, line 42: This would be a good oportunity to use `Registry Name` instead.

glowredman (commented):
One enhancement could be to use ChatComponents such that, when clicking on any of the values, that values is copied to the clipboard. The current behaviour (MT snippet being copied to clipboard) should remain though.

In any case, this is a practical addition and long overdue.

PR #86, opened 2024-10-26
Add texture debug feature

Alexdoru:
Using `-Dgtnhlib.debugtextures=true` will now create two files in your `.minecraft` folder that will list all the registered textures in a `csv` file with their name, width, height, amount of frames and the size in bytes it takes in the RAM.

We should use this to point to all the textures that are too large and rework them.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
@@ -82,7 +82,12 @@ public class GTNHLibCore extends DummyModContainer implements IFMLLoadingPlugin,
 
     @Override
     public List<String> getMixins(Set<String> loadedCoreMods) {
-        return Mixins.getEarlyMixins(loadedCoreMods);
+        final List<String> mixins = Mixins.getEarlyMixins(loadedCoreMods);
+        if (Boolean.parseBoolean(System.getProperty("gtnhlib.debugtextures", "false"))) {
+            mixins.add("debug.MixinDynamicTexture");
+            mixins.add("debug.MixinTextureAtlasSprite");
+        }
+        return mixins;
     }
 
     @Override
--- a/src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
@@ -82,12 +82,7 @@ public class GTNHLibCore extends DummyModContainer implements IFMLLoadingPlugin,
 
     @Override
     public List<String> getMixins(Set<String> loadedCoreMods) {
-        final List<String> mixins = Mixins.getEarlyMixins(loadedCoreMods);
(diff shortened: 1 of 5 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
```diff
@@ -82,7 +82,12 @@ public String getMixinConfig() {
 
     @Override
     public List<String> getMixins(Set<String> loadedCoreMods) {
-        return Mixins.getEarlyMixins(loadedCoreMods);
+        final List<String> mixins = Mixins.getEarlyMixins(loadedCoreMods);
+        if (Boolean.parseBoolean(System.getProperty("gtnhlib.debugtextures", "false"))) {
```
Lyfts, line 86: That should be in the Mixins enum class like the rest of them with an apply condition of `.setApplyIf(() -> Boolean.getBoolean("gtnhlib.debugtextures"))`

Lyfts (changes requested):

src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
Alexdoru: why ? it's a debug feature

Lyfts: Why would that matter, either way it only gets applied if the jvm flag is set

src/main/java/com/gtnewhorizon/gtnhlib/mixins/Mixins.java
```diff
@@ -18,12 +18,14 @@ public enum Mixins {
...
-
     GUI_MOD_LIST(new Builder("Auto config ui").addTargetedMod(TargetedMod.VANILLA).setSide(Side.CLIENT)
             .setPhase(Phase.EARLY).addMixinClasses("fml.MixinGuiModList")),
-
     EVENT_BUS_ACCESSOR(new Builder("EventBusAccessor").addTargetedMod(TargetedMod.VANILLA).setSide(Side.BOTH)
-            .setPhase(Phase.EARLY).addMixinClasses("fml.EventBusAccessor", "fml.EnumHolderAccessor"));
+            .setPhase(Phase.EARLY).addMixinClasses("fml.EventBusAccessor", "fml.EnumHolderAccessor")),
+    DEBUG_TEXTURES(new Builder("Dump textures sizes").addTargetedMod(TargetedMod.VANILLA).setSide(Side.CLIENT)
+            .setPhase(Phase.EARLY)
+            .setApplyIf(() -> Boolean.parseBoolean(System.getProperty("gtnhlib.debugtextures", "false")))
```
Lyfts, line 27: nit: `Boolean.parseBoolean(System.getProperty("gtnhlib.debugtextures", "false")` can be shortened to just `Boolean.getBoolean("gtnhlib.debugtextures")`

Lyfts (approved):

PR #88, opened 2024-11-11
Disable thread safety checks with FalseTweaks

ah-OOG-ah:
If FT's threaded rendering is running, it's handling the safety checks itself so skip ours.

### DIFF
--- a/repositories.gradle
+++ b/repositories.gradle
@@ -1,5 +1,7 @@
 // Add any additional repositories for your dependencies here
 
 repositories {
-
+    maven {
+        url = uri("https://mvn.falsepattern.com/releases")
+    }
 }
--- a/repositories.gradle
+++ b/repositories.gradle
@@ -2,6 +2,10 @@
 
 repositories {
     maven {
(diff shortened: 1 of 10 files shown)


FalsePattern:
`TessellatorManager.get()` should also be exposed in the API, see: https://github.com/FalsePattern/FalseTweaks/pull/197/commits/62345907099b0f2c962df8e7d0f14f0b264a34b1#diff-e17dd99acbec90e61fb1cbc3633acbb173a690333d04dab26a0c36f80a53d1daR523-R525

repositories.gradle
```diff
@@ -1,5 +1,7 @@
 // Add any additional repositories for your dependencies here
 
 repositories {
-
+    maven {
+        url = uri("https://mvn.falsepattern.com/releases")
```
mitchej123, line 5: Limit this, similar to how we do with other repositories.   Example:
```
    maven {
        name = 'taumc'
        url = 'https://maven.taumc.org/releases'
        content {
            includeGroup 'org.taumc'
        }
    }
```

mitchej123 (approved):

PR #94, opened 2024-11-26
Pull LWJGL3 backports to GTNHLib

ah-OOG-ah:
While these used to be in Angelica, I noticed at least one other mod (Avaritia) doing problematic amounts of direct buffer allocation and I don't want to slap Angelica deps on things.


ah-OOG-ah:
Looking into fixing the largest issue in Angelica 2.x, so this is likely uneeded

PR #102, opened 2025-01-19
Easily sync config values between server & client

Lyfts:
This adds a way to easily sync config values from server -> client by simple annotating the config field (or class) with `@Config.Sync`.
All values are restored to their original value once the player leaves the server.

Tested with this lovely config https://github.com/GTNewHorizons/GTNHLib/blob/7b65533a67926446c2e98f77c4503578dd049226/src/main/java/com/gtnewhorizon/gtnhlib/config/TestConfig.java and everything seems to sync correctly.


serenibyss (approved):
Great work, this should be a very useful feature

mitchej123:
Leaving this comment here for any future crashes that come out of this:

I like the idea overall, but I think that syncing MC restart fields is a bad idea.  A lot of configs use this for mixins, and also gate access to mixins introduced fields/changes later.    Without an annotation processor to warn at compile time this leaves a lot of room open for immediate client side crashes if someone marks such a config as Sync without realizing the problems.

PR #104, opened 2025-01-23
Synced Keybinds API

serenibyss:
Creates a new API for syncing player client key presses to the server for specified keys.

Supports two different types of key checks:
- "Key Held" - If the key is currently being held down by the player
- "Key Pressed" - Fired for a registered set of listeners when the key is pressed

Supports lightweight keybinds for a `Keyboard` key code, wrapped MC `KeyBinding` types with server syncing, and brand new `KeyBinding` binds with configuration available on the controls page like normal

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/keybind/SyncedKeybind.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/keybind/SyncedKeybind.java
@@ -55,7 +55,8 @@ public final class SyncedKeybind {
     private boolean isKeyDown;
 
     private final WeakHashMap<EntityPlayerMP, Boolean> mapping = new WeakHashMap<>();
-    private final WeakHashMap<EntityPlayerMP, Set<IKeyPressedListener>> listeners = new WeakHashMap<>();
+    private final WeakHashMap<EntityPlayerMP, Set<IKeyPressedListener>> playerListeners = new WeakHashMap<>();
+    private final Set<IKeyPressedListener> globalListeners = Collections.newSetFromMap(new WeakHashMap<>());
 
     // Doubly-wrapped supplier for client-side only type
     private SyncedKeybind(Supplier<Supplier<KeyBinding>> keybindingGetter) {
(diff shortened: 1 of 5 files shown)


Caedis:
How will this interact with https://github.com/Nova-Committee/ModernKeyBinding/tree/Forge/1.7.10 ?
Have not gotten around to merging it into our Controlling fork

Caedis:
Because you are calling, keybind.isKeyPressed, it should be fine

Lyfts (commented):
Is there a reason why we would need to register a listener for every player per keybind? 
I can think of loads of keybinds that only exist to send a packet to the server so it would be nice to have the option of always doing an action server-side when a player presses the keybind. (registering a Consumer<EntityPlayer> that gets called when it's pressed maybe?)

serenibyss:
> Is there a reason why we would need to register a listener for every player per keybind? I can think of loads of keybinds that only exist to send a packet to the server so it would be nice to have the option of always doing an action server-side when a player presses the keybind. ~(registering a Consumer that gets called when it's pressed maybe?)~ Scratch that, just a global listener with the existing interface would be great

Originally when I wrote this code (quite some time ago for another project), it was written primarily for armor keybinds, so the armor item could subscribe for a player in that case. I think there's merit to both ways of listening though, so I'll look into supporting either way

src/main/java/com/gtnewhorizon/gtnhlib/keybind/SyncedKeybind.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * Registers an {@link IKeyPressedListener} to this key, which will have its {@link IKeyPressedListener#onKeyPressed
+     * onKeyPressed} method called when any player presses this key.
+     *
+     * @param listener The handler for the key clicked event.
+     */
+    public void registerGlobalListener(IKeyPressedListener listener) {
+        globalListeners.add(listener);
```
Lyfts, line 165: nit: it would be nice if this returned the SyncedKeybind so we could chain it in a static final field.

Lyfts (approved):
Tested and works well, great work!

src/main/java/com/gtnewhorizon/gtnhlib/keybind/SyncedKeybind.java
Caedis: Surprised maya didnt make this a builder

PR #112, opened 2025-02-06
Introduce Capability system

miozune:
This PR introduces Capability system.

### For those who don't know about Capability

Please read Javadoc. If there's anything unclear, feel free to ask questions.

### For those who are familiar with Capability in later MC versions

This PR was originally based on 1.12.2 Forge implementation. After that Raven's suggestion drastically simplified it. Now `Capability` class is gone and there're only `CapabilityProvider` and `Capabilities`.

Side note, due to skill issue I used generative AI to generate documents. While I used multiple models, read everything they outputted and tweaked sentences, there could still be oddities.


eigenraven (changes requested):
The docs definitely need rewording as they're quite confusing in places, I'll try to do it later today

src/main/java/com/gtnewhorizon/gtnhlib/capability/ICapabilityProvider.java
```diff
@@ -0,0 +1,51 @@
...
+ *             if (capability == MyCapability.CAPABILITY) {
+ *                 return MyCapability.CAPABILITY.cast(myCapability);
+ *             }
+ *             return null;
+ *         }
+ *     }
+ * }</pre>
+ */
+// spotless:on
+public interface ICapabilityProvider {
```
YannickMG, line 31: This is worded pretty generically but as written this only makes sense for tile entities, which have sides. I would like to use capabilities for Covers and probably other things which might not have sides at a later time. Should they all be burdened with the concept of side? Should this be `ISidedCapabilityProvider` while `ICapabilityProvider` wouldn't have sides?

Perhaps not but I'd like to hear to what extent capabilities for non-tile-entities are planned to be supported.

In fact, could we wait to merge this until this includes an example of a capability for something  other than a tile entity?

src/main/java/com/gtnewhorizon/gtnhlib/capability/CapabilityUtil.java
```diff
@@ -0,0 +1,111 @@
...
+     * Retrieves a capability from the given TileEntity.
+     *
+     * @param tileEntity The TileEntity to query.
+     * @param capability The {@link Capability} being requested.
+     * @param side       The side of the TileEntity being queried.
+     * @param <T>        The type of the capability interface.
+     * @return The capability implementation, or null if not available.
+     */
+    public static <T> T getCapability(@Nullable TileEntity tileEntity, @Nonnull Capability<T> capability,
+            @Nullable ForgeDirection side) {
```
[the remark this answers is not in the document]
miozune: Good catch, I forgot it exists

src/main/java/com/gtnewhorizon/gtnhlib/capability/ICapabilityProvider.java
miozune: `side` can be `UNKNOWN` if 1. Queried capability doesn't have concept of sidedness 2. Queried capability is about TileEntity, but sidedness is not the interest. It's technically possible to separate into the two, but having two methods for providing capability sounds confusing to me.
Actually I tried to implement generics "context" rather than just `ForgeDirection`, just like seen in NeoForge. However, I found it difficult to implement cleanly due to the difference of how capability is registered. Moreover, I skimmed through popular repositories and found only `Void` and `Direction` being used, meaning generic context is useless, at least at this time.

Caedis:
Some examples like for some vanilla stuff would be useful

src/main/java/com/gtnewhorizon/gtnhlib/capability/CapabilityUtil.java
```diff
@@ -0,0 +1,111 @@
...
+     * Retrieves a capability from the given Entity.
+     *
+     * @param entity     The Entity to query.
+     * @param capability The {@link Capability} being requested.
+     * @param <T>        The type of the capability interface.
+     * @return The capability implementation, or null if not available.
+     */
+    public static <T> T getCapability(@Nullable Entity entity, @NotNull Capability<T> capability) {
+        return getCapability(entity, capability, ForgeDirection.UNKNOWN);
+    }
```
YannickMG, line 85: Why does this only accept an Entity while the other override accepts an Object & a Side? This means if I want to get a capability of a non-entity, a side is required?

miozune:
Now it's reworked and significantly simplified thanks to Raven. Please take another look.

> Some examples like for some vanilla stuff would be useful

As written on Javadoc, usage of vanilla/Forge interfaces as capability is discouraged.

src/main/java/com/gtnewhorizon/gtnhlib/capability/CapabilityUtil.java
miozune: Method accepting `Object` is set to private. Actually anything can work with capability, but I want to ensure type safety and listed things that one might want to use capability in MC modding context.

eigenraven (approved):

YannickMG (approved):

PR #116, opened 2025-02-24
Notify client of servers view distance

Lyfts:
I have a very niche use case for this in a certain client-only mod and calculating it client-side gets extremely messy.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
@@ -1,15 +1,18 @@
 package com.gtnewhorizon.gtnhlib;
 
 import net.minecraft.entity.player.EntityPlayerMP;
+import net.minecraft.server.MinecraftServer;
 import net.minecraft.util.IChatComponent;
 import net.minecraftforge.common.util.FakePlayer;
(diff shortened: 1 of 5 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
```diff
@@ -86,4 +92,11 @@ public void sendMessageAboveHotbar(EntityPlayerMP player, IChatComponent chatCom
         NetworkHandler.instance
                 .sendTo(new PacketMessageAboveHotbar(chatComponent, displayDuration, drawShadow, shouldFade), player);
     }
+
+    @SubscribeEvent
+    public static void onPlayerLogin(PlayerEvent.PlayerLoggedInEvent event) {
+        if (!(event.player instanceof EntityPlayerMP playerMP)) return;
+        int distance = MinecraftServer.getServer().getConfigurationManager().getViewDistance();
+        NetworkHandler.instance.sendTo(new PacketViewDistance(distance), playerMP);
+    }
```
serenibyss, line 101: Won't this be inaccurate if it's changed after login, since it is only updated on logged in event?

serenibyss (approved):

PR #124, opened 2025-03-20
Make BlockPos modifications return BlockPos instead of IBlockPos

Taskeren:
This PR replaces the return type of methods in BlockPos, but not changing the IBlockPos interface.
Idk if it ABI compat, but it does API compat.

This fixes the problem that you'll need to cast the return values to BlockPos, when you maintaining a Collection of BlockPos.
In newer forges/fabrics (e.g.: 1.12.2), there isn't a IBlockPos interface, so all methods return BlockPos. It is misleading.

And one more reason is that, BlockPos is _mutable_, so at least the return types should be `IMutableBlockPos` instead of `IBlockPos`, because it loses the set methods.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/blockpos/BlockPos.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/blockpos/BlockPos.java
@@ -40,17 +40,17 @@ public class BlockPos extends Vector3i implements IMutableBlockPos {
     }
 
     @Override
-    public IBlockPos offset(ForgeDirection d) {
+    public BlockPos offset(ForgeDirection d) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/blockpos/BlockPos.java
```diff
@@ -40,17 +40,17 @@ public int getZ() {
     }
 
     @Override
-    public IBlockPos offset(ForgeDirection d) {
```
mitchej123, line 43: But why?

mitchej123 (changes requested):

mitchej123 (approved):

PR #126, opened 2025-03-27
Backport callFromMainThread() and addScheduledTask()

Taskeren:
Backport callFromMainThread() and addScheduledTask() of MinecraftServer from 1.12.2

The executing timing is exactly same to what 1.12 Forge does.
And because of my Mixin knowledge limit, I can only get the server thread like that.

These functions are used to do something in server thread, submitted from other threads, like async tasks. (although it checks if the calling thread is the server thread, like 1.12 does)

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/util/ServerThreadUtil.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/util/ServerThreadUtil.java
@@ -47,6 +47,13 @@ public class ServerThreadUtil {
         ServerThreadUtil.serverThread = serverThread;
     }
 
+    @ApiStatus.Internal
+    public static void clear() {
+        futureTaskQueue.clear();
+        ServerThreadUtil.server = null;
+        ServerThreadUtil.serverThread = null;
+    }
+
     public static boolean isCallingFromMinecraftThread() {
         return Thread.currentThread() == serverThread;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/util/ServerThreadUtil.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListenableFutureTask;
+
+@SuppressWarnings("UnstableApiUsage")
+public class ServerThreadUtil {
+
+    private static final Logger logger = LogManager.getLogger();
+    private static final Queue<FutureTask<?>> futureTaskQueue = Queues.newArrayDeque();
+    private static MinecraftServer server;
```
Lyfts, line 26: This is never nullified when the server exits and will cause the IntegratedServer instance to leak when going from SP->MP

Lyfts (dismissed):

src/main/java/com/gtnewhorizon/gtnhlib/util/ServerThreadUtil.java
Taskeren: Fixed.

Glease (approved):

Glease (approved):

PR #127, opened 2025-04-01
New mixin infrastructure

Cleptomania:
This re-works the mixin infrastructure to be more broadly usable by all mods, so that mods don't need to copy/paste all of this into themselves and can just pull it from GTNHLib. As of right now this is a first pass at what it might look like, looking for comments/ideas on this still.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/core/GTNHLibCore.java
@@ -4,7 +4,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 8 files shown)


eigenraven (approved):

Cleptomania:
The tldr on this PR is mods make a Mixins enum much like before, but it implements the new IMixins interface(methods provided by lombok getters in this particular case).  They can optionally provide their own `TargetedMod` enum implementing that interface if GTNHLib default enum doesn't have the things they need. 

Then you can call `IMixins.getEarlyMixins(Mixins.class, loadedCoreMods)` for example in your coremod to get the early mixin list. Similar for late mixins. In that example, `Mixins.class` is referring to the mod's enum that implements IMixins.

eigenraven (approved):

PR #138, opened 2025-05-26
Update SpatialHashGrid findNearby

Caedis:
Changed default formula to Chebyshev
Added test

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/datastructs/spatialhashgrid/SpatialHashGrid.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/datastructs/spatialhashgrid/SpatialHashGrid.java
@@ -101,7 +101,7 @@ public class SpatialHashGrid<T> {
 
         // Make sure that cells which partially fall in the radius are still checked
         final int cellRad = (radius + cellSize - 1) / cellSize;
-        final int radiusSquared = radius * radius;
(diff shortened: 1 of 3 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/datastructs/spatialhashgrid/SpatialHashGrid.java
```diff
@@ -83,7 +83,7 @@ public void remove(T obj) {
      * @return list of nearby objects
      */
     public List<T> findNearby(int x, int y, int z, int radius) {
-        return findNearby(x, y, z, radius, DistanceFormula.SquaredEuclidean);
+        return findNearby(x, y, z, radius, DistanceFormula.Chebyshev);
```
YannickMG, line 86: You didn't update the JavaDoc.

### Not part of a review, just a side comment:
I will mention that if you have to read the JavaDoc to understand which metric is being used, maybe the metric should be in the name.

The override for the other 2 methods isn't expected to be called with all 3 arguments given one would be wordier and there is no use case for calling it with an arbitrary number of distance methods, you always know which one you want. Given that, it might be simpler to just have 3 appropriately named methods.

Caedis: Is that what you were wanting?

YannickMG (approved):

PR #140, opened 2025-05-31
add close() to ShaderProgram to free resources

MeiTianyou:
does what it says on the tin


src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/shader/ShaderProgram.java
```diff
@@ -12,6 +13,7 @@
 import org.lwjgl.opengl.GL20;
 
 import com.gtnewhorizon.gtnhlib.GTNHLib;
+import com.gtnewhorizon.gtnhlib.bytebuf.MemoryStack;
 
 @SuppressWarnings("unused")
 public class ShaderProgram {
```
eigenraven, line 19: Can you implement AutoCloseable as that lets you use shaders in a try-with-resources block? I don't think it will be a very common usecase, but it's a good convention to follow for classes with close() methods

eigenraven (approved):

PR #146, opened 2025-06-06
add VolumeMembershipCheck

Alexdoru:
This is a data structure that allows you to quickly check if an arbitrary point in space is contained within at least one of the volumes defined.

It does so by brute forcing through all the positions.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/datastructs/space/VolumeMembershipCheck.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/datastructs/space/VolumeMembershipCheck.java
@@ -38,13 +38,13 @@ public class VolumeMembershipCheck {
         return null;
     }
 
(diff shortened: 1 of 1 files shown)


Taskeren:
In DimensionData, is that you are manaing data in a array where they are separated into 4 slots? I think it would be better in code if you just make a data class and store the data class instance in the array.

And I also believe the DimensionData can use delegate design to use builtin data structures like ArrayList, so that you don't have to handle things so "basic".

Alexdoru:
> In DimensionData, is that you are manaing data in a array where they are separated into 4 slots? I think it would be better in code if you just make a data class and store the data class instance in the array.

The goal is speed, looping over a single array of primitives is faster than looping over an array of objects that contains the pritimes. Morever it uses less space to represent in memory.

Caedis (approved):

PR #149, opened 2025-06-19
Add Brigadier

Taskeren:
This PR adds the limited Brigadier support (the modern Minecraft command system) to current legacy command system.

It now can be properly executed, listed in the help command, and provides suggestions, `/help <brigadier-command-name>`.

~~More argument types provided in modern Minecraft are not yet included, but maybe I will add them in a future PR.~~

~~I don't have a plan to add more things like modern GuiNewChat that supports brigadier suggestions now.~~

The brigadier library is licensed under MIT, so I think it's ok to simply shadow it in GTNHLib.

After asking in the discord, backporting things from modern MC may face legal problems, so I probably won't do any of them.
But it is still needed to have something like `BlockPosArgumentType` and other powerful argument types. Help wanted.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
@@ -6,9 +6,8 @@ import net.minecraft.server.MinecraftServer;
 import net.minecraft.util.IChatComponent;
 import net.minecraftforge.common.util.FakePlayer;
 
-import com.gtnewhorizon.gtnhlib.brigadier.BrigadierApi;
-import com.gtnewhorizon.gtnhlib.config.ConfigException;
 import com.gtnewhorizon.gtnhlib.block.BlockTest;
+import com.gtnewhorizon.gtnhlib.brigadier.BrigadierApi;
 import com.gtnewhorizon.gtnhlib.config.ConfigurationManager;
 import com.gtnewhorizon.gtnhlib.eventbus.AutoEventBus;
 import com.gtnewhorizon.gtnhlib.eventbus.EventBusSubscriber;


Speiger:
@Dream-Master always have been.
But reduced my legacy mc version time drastically :)
But you guys know if bugs need to be fixed i am present xD

Speiger:
Update: https://github.com/Speiger/Salutation

Speiger:
@Taskeren getting brigadier fully running isn't an option especially since i maintain the full 1.7.10-1.12.2 range with chunkpregen.
This mod will not collide with brigadier as it is a custom implementation that just implements a vanilla command and gives you  the 1.19.2 structure on how commands are written with the limitations of vanillla <1.12.2.
Which you are bound to as well.

In other words they should be able to coexist.
Also FYI, all the code that i stripped out of my brigadier implementation is effectively all the features you can't support anyways.
On top of that i got rid of a ton of boiler plate code where you need to make huge lambda chains when the code can be KISS and clean.
And i have QoL functions that mojang is aware of since 2 years at this point and hasn't implemented at all.
Stuff like: "getOrDefault" or "has" are feature still missing in brigadier which require you to make wrapper functions that catch exceptions because mojang doesn't need those so why would modders anyways? Just write more versions of the same functions.


So no i won't support brigadier directly as it is outside of the scope of the project, nor am i interested, its just an optional feature list that i wrote because i dislike the 1.7.10 command system over the 1.19.2 command system and 1.7.10 doesn't support all that 1.19.2 supports anyways :)

Sorry if that comes over as a rant ^^" I don't mean any harm and if i did then i apologize for doing so.

RecursivePineapple (approved):

PR #154, opened 2025-07-01
Add support for mouse buttons in SyncedKeybinds

Caedis:


src/main/java/com/gtnewhorizon/gtnhlib/keybind/SyncedKeybind.java
```diff
@@ -221,6 +222,18 @@ void updateKeyDown(boolean keyDown, EntityPlayerMP player) {
...
+
+    // separate from the keyboard handler to prevent mouse movement triggers when holding a key
+    @SubscribeEvent
+    @SideOnly(Side.CLIENT)
+    public static void onInputEvent(InputEvent.MouseInputEvent event) {
+        if (!Mouse.getEventButtonState()) return;
+        onInputEvent();
+    }
+
+    private static void onInputEvent() {
```
serenibyss, line 236: I would mark this as client side only just to be safe

PR #156, opened 2025-07-04
Remove key pressed packet

Caedis:
Only send a packet to the server on key down and up instead of every tick a key is pressed. Server can check if a key is held down by calling isKeyDown and passing the player.
onKeyPressed updated to take isDown state

AE2 stuff does not use a server listener so it doesnt need updating.
Backhand does and I will update that.


Alexdoru:
shouldn't the `SyncedKeybind` class be split into 2 ? the keybind itself and a keybindmanager class? would make things more understandable imo

Caedis:
Its pretty understandable to me since its all logic besides the packets are in the same place

serenibyss (approved):

PR #157, opened 2025-07-04
add ArrayProximityCheck4D and ArrayProximityMap4D<T>

Alexdoru:
add a map version of the proximity check that keeps a reference to the object mapped . Using this is faster than spatial hash grid
```
Benchmark                   Mode      Score     Error  Units
baseRandom                  avgt     28,543 +-  0,002  ns/op
arrayGetClosest_size20      avgt    100,342 +-  0,388  ns/op
hashGridGetClosest_size20   avgt  10198,580 +- 75,006  ns/op
arrayGetClosest_size100     avgt    392,217 +-  0,218  ns/op
hashGridGetClosest_size100  avgt   6904,390 +- 35,042  ns/op
arrayGetClosest_size500     avgt   1857,026 +-  2,160  ns/op
hashGridGetClosest_size500  avgt   8006,260 +- 82,044  ns/op
```

Needs to be merged with https://github.com/GTNewHorizons/GT5-Unofficial/pull/4549

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/datastructs/space/ArrayProximityMap4D.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/datastructs/space/ArrayProximityMap4D.java
@@ -9,8 +9,9 @@ import java.util.function.Consumer;
 import javax.annotation.Nonnull;
 
 /**
- * This is a data structure that allows you to quickly check if an arbitrary point in space (dim, x, y, z) is contained
- * within at least one of the volumes defined. You can also retrieve the closest stored object from that point if any.
+ * This is a data structure that allows you to quickly check if an arbitrary point in space (dim, x, y, z) is in range
+ * of at least one of the objects defined. You can also retrieve the closest stored object from that point if any. Or
+ * perform an action for each object in range.
  * <p>
  * When adding an object to an existing position, it won't add a second object but will instead replace the element at
  * this position and update the radius.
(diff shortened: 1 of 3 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/datastructs/space/ArrayProximityMap4D.java
```diff
@@ -0,0 +1,307 @@
...
+ * This is a data structure that allows you to quickly check if an arbitrary point in space (dim, x, y, z) is contained
+ * within at least one of the volumes defined. You can also retrieve the closest stored object from that point if any.
+ * <p>
+ * When adding an object to an existing position, it won't add a second object but will instead replace the element at
+ * this position and update the radius.
+ */
+@SuppressWarnings("unused")
+public class ArrayProximityMap4D<T> {
+
+    // optimization idea : this can be dynamically
```
RecursivePineapple, line 20: You can swap this to a Int2ObjectMap, then swap between Int2ObjectOpenHashMap and Int2ObjectArrayMap at runtime as needed

Alexdoru: i know, but this is used for things like maglev, mob repellors, wireless chargers and you're probably not gonna have some in more than 3 dims lol

Caedis:
Do benchmarks with counts 5, 20, 50, 200, 500, 1000, 2500

Doing just a 5 count benchmark will obviously show the arrays being a lot faster

Group the items into clumps of random amounts and scatter them across the map to simulate bases.

For each clump, do contain iterations of varying ranges

Caedis (approved):

PR #160, opened 2025-07-10
Add Cyrillic support for MathExpressionParser

Eldrinn-Elantey:


Eldrinn-Elantey:
Yeah, I checked with `./gradlew runClient'. The client did not crash when using

chrombread (approved):

PR #167, opened 2025-09-15
Add Generic Item I/O Capabilities

RecursivePineapple:
I'm using these in GT to improve the item moving code, but I'd like to move the AE integration into AE instead of putting it in GT and hacking around AE's limitations. By doing this I'll be able to return the sink and source implementations from AE via capabilities, which will allow me to remove the special edge cases in GT and decouple the two completely.

This doesn't need to go into 2.8 because the feature isn't close to being ready, but I doubt dream wants more work for the experimentals so I marked this as safe for stable. This shouldn't cause any issues because it isn't actually being called anywhere, it's just a set of interfaces.


src/main/java/com/gtnewhorizon/gtnhlib/capability/item/AbstractInventorySourceIterator.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        return extracted;
+    }
+
+    @Override
+    public void insert(ItemStack stack) {
+        if (!isStackValid(stack)) return;
+
+        ItemStack inSlot = getStackInSlot(slots[last]);
```
DarkShadow44: I'm very confused by that `last` thing. Can't we just have `current` instead of two indices?

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/InventoryItemSink.java
```diff
@@ -0,0 +1,173 @@
...
+
+    private BitSet allowedSlots = null;
+
+    public InventoryItemSink(IInventory inv, ForgeDirection side) {
+        this.inv = inv;
+        this.side = side;
+    }
+
+    @Override
+    public void setAllowedSinkSlots(int @org.jetbrains.annotations.Nullable [] slots) {
```
DarkShadow44: Why no import for `Nullable` here?

src/main/java/com/gtnewhorizon/gtnhlib/util/data/ItemId.java
```diff
@@ -0,0 +1,132 @@
...
+        return new ItemId(stack.getItem(), OreDictionary.WILDCARD_VALUE, null);
+    }
+
+    public static ItemId createAsWildcardWithNBT(ItemStack stack) {
+        return create(stack.getItem(), OreDictionary.WILDCARD_VALUE, stack.getTagCompound());
+    }
+
+    public static ItemId createWithoutNBT(ItemStack stack) {
+        return new ItemId(stack.getItem(), Items.feather.getDamage(stack), null);
+    }
```
DarkShadow44: Why does createAsWilcard default to "without NBT" but "create" default to "with NBT"? It's a bit confusing. Maybe only have WithNBT and WithoutNBT methods? Then it can't accidentally go wrong.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/InventoryItemSink.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    @Override
+    public ItemStack store(ItemStack stack) {
+        stack = ItemStack.copyItemStack(stack);
+
+        if (stack.stackSize <= 0) return null;
+
+        this.markedDirty = false;
```
DarkShadow44: Why does storing mark it as not dirty?

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/InventoryItemSink.java
```diff
@@ -0,0 +1,173 @@
...
+            ItemStack inserted = ItemUtil.copyAmount(toTransfer, stack);
+            inv.setInventorySlotContents(slot, inserted);
+            stack.stackSize -= toTransfer;
+
+            markDirty();
+
+            if (stack.stackSize <= 0) return null;
+        }
+
+        return stack.stackSize <= 0 ? null : stack;
```
DarkShadow44: Again, should be enough to just return stack, should not be possible to be 0 amount here.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/AbstractInventorySourceIterator.java
```diff
@@ -0,0 +1,159 @@
...
+
+    protected final int getCurrentSlot() {
+        return slots[last];
+    }
+
+    public static boolean isStackValid(ItemStack stack) {
+        return stack != null && stack.getItem() != null && stack.stackSize > 0;
+    }
+
+    public static boolean areStacksEqual(ItemStack a, ItemStack b) {
```
DarkShadow44: Isn't this the same as in ItemUtils?

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/AbstractInventorySourceIterator.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        setInventorySlotContents(slots[last], out);
+    }
+
+    protected final int getCurrentSlot() {
+        return slots[last];
+    }
+
+    public static boolean isStackValid(ItemStack stack) {
```
DarkShadow44: Can't you use the one from ItemUtils?

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/ItemTransfer.java
```diff
@@ -0,0 +1,179 @@
...
+
+        if (source == null) return 0;
+        if (sink == null) return 0;
+        if (stacksToTransfer == 0) return 0;
+        if (maxItemsPerTransfer == 0) return 0;
+
+        source.setAllowedSourceSlots(sourceSlots);
+        sink.setAllowedSinkSlots(sinkSlots);
+
+        InventorySourceIterator iter = source.iterator();
```
DarkShadow44: Why does this function use `InventorySourceIterator`? Shouldn't the Source/Sink be enough? The sink does so much work finding the correct slots to put stuff in, why can't  the source have a single extract method that finds the next matching itemstack?

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/IItemSink.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    ItemStack store(ItemStack stack);
+
+    /**
+     * Sets the slots that this sink can insert into. If any given slot (or all of them) are invalid, they may be
+     * ignored. The sink must never crash, delete items, etc if a slot is invalid - it must be silently ignored.
+     * <p />
+     * If the sink does not have something like a 'slot' (i.e. an ME system), the given slots can be ignored. If the
+     * sink does have slots, but the given slots are invalid (i.e. a single chest with the filter set to slot 100),
+     * items must never be inserted.
```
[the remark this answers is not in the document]
RecursivePineapple: You wouldn't, this is for stuff like GT's robot arm where the player can filter for specific slots by index. The indices are just magic numbers that aren't specified anywhere.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/InventoryItemSink.java
RecursivePineapple: Looks like intellij imported two different `@Nullable`s, I'll go through everything and fix them.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/AbstractInventorySourceIterator.java
```diff
@@ -0,0 +1,146 @@
...
+     */
+    protected abstract void setInventorySlotContents(int slot, ItemStack stack);
+
+    /**
+     * Marks the backing inventory dirty. Only called the first time a stack is set.
+     * @see IInventory#markDirty()
+     */
+    protected void markDirty() {
+
+    }
```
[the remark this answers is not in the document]
DarkShadow44: I don't see why that means that `dirty = true` can't be moved into markDirty though? It shouldn't change the logic. Just that calling setDirty from the outside also sets the flag.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/InventorySourceIterator.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Extracts items from the current index. May return less items than what was originally reported for this index.
+     *
+     * @param amount The amount to extract
+     * @return The extracted stack
+     */
+    ItemStack extract(int amount);
+
+    /**
+     * Returns the rejected items from a previously extracted stack. The item must be identical in every way, except the
```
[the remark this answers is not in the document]
DarkShadow44: Wouldn't `putBackRejected` or so be less confusing then?

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/InventoryItemSink.java
```diff
@@ -0,0 +1,173 @@
...
+            inSlot.stackSize += toTransfer;
+
+            inv.setInventorySlotContents(slot, inSlot);
+
+            markDirty();
+
+            if (stack.stackSize <= 0) return null;
+        }
+
+        if (stack.stackSize <= 0) return null;
```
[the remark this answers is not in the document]
RecursivePineapple: It is, line 97 decrements the stackSize. If it's >0 then it means there are items that need to be inserted into empty slots.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/AbstractInventorySourceIterator.java
RecursivePineapple: Yeah, looks like I forgot to remove these after I added ItemUtil. I didn't want to add a util class to gtnhlib but I had to unless I wanted to copy these methods everywhere.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/IItemSink.java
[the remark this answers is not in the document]
RecursivePineapple: Yeah, that's generally how it works. The sink will only insert into slots it's allowed to. It fills the slots it has access to and rejects any items it couldn't fit. If it was told to insert into invalid slots it rejects everything.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/AbstractInventorySourceIterator.java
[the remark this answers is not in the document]
RecursivePineapple: I moved it and added a must-call-super annotation

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/InventoryItemSink.java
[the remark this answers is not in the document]
DarkShadow44: When we enter that loop at line 71 it's guaranteed that stacksize is > 0. If the decrement in line 97 makes it smaller than 0, then line 104 returns already. I still think the check in line 107 is unneeded.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/InventorySourceIterator.java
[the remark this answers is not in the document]
RecursivePineapple: I changed it to `reinsertRejected`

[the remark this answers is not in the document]
RecursivePineapple: I changed it back to `insert`, but in my defence the iterator is generic now :stuck_out_tongue:

src/main/java/com/gtnewhorizon/gtnhlib/item/AbstractInventoryIterator.java
```diff
@@ -0,0 +1,187 @@
...
+
+        if (!ItemUtil.isStackEmpty(inSlot) && !ItemUtil.areStacksEqual(inSlot, stack)) {
+            return stack;
+        }
+
+        if (!canInsert(stack, slotIndex)) {
+            return stack;
+        }
+
+        if (!ItemUtil.isStackEmpty(inSlot)) {
```
DarkShadow44: I'd move the stack copy before here, so we don't copy in early exit scenarios

src/main/java/com/gtnewhorizon/gtnhlib/item/ItemTransfer.java
```diff
@@ -0,0 +1,201 @@
...
+        sink.setAllowedSinkSlots(sinkSlots);
+
+        InventoryIterator iter = source.sourceIterator();
+
+        // Don't bother supporting iterator-less sources, it'll just be a performance and logic nightmare
+        if (iter == null) return 0;
+
+        int itemsTransferred = 0, stacksTransferred = 0;
+
+        outer: while (iter.hasNext() && stacksTransferred < stacksToTransfer
```
DarkShadow44: Why not use source.pull here? Seems much easier to not use the iterator here but just pull and push. Although we should probably make sure to not always recreate the sinks iterator...

src/main/java/com/gtnewhorizon/gtnhlib/item/InventoryItemSink.java
```diff
@@ -0,0 +1,121 @@
...
+
+        return stack;
+    }
+
+    @Override
+    public @NotNull StandardInventoryIterator sinkIterator() {
+        return new StandardInventoryIterator(inv, side, slots, allowedSlots) {
+
+            @Override
+            protected int getSlotStackLimit(int slot, ItemStack stack) {
```
DarkShadow44: Why do  we need to override that method? Seems like it does the same as the one in StandardInventoryIterator

src/main/java/com/gtnewhorizon/gtnhlib/item/ItemStackPredicate.java
```diff
@@ -0,0 +1,57 @@
...
+ * A predicate for ItemStacks.
+ */
+@FunctionalInterface
+public interface ItemStackPredicate extends Predicate<ItemStack> {
+
+    @Override
+    boolean test(ItemStack stack);
+
+    default boolean test(ImmutableItemStack stack) {
+        return test(stack.toStack());
```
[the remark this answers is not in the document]
RecursivePineapple: It's meant to be overridden if possible. This isn't an ideal implementation, but I wanted this predicate to still be a `Predicate<ItemStack>` functional interface so that you could use lambdas with it.

src/main/java/com/gtnewhorizon/gtnhlib/capability/item/ItemSink.java
```diff
@@ -0,0 +1,114 @@
...
+    default void resetSink() {
+
+    }
+
+    /**
+     * Injects a stack into this sink. This operation is not atomic. Any rejected items are returned. There is no
+     * guarantee that the sink is completely full when items are rejected, this is just a best-effort operation. The
+     * stack parameter must not be mutated by the implementation.
+     */
+    ItemStack store(ItemStack stack);
```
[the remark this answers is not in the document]
RecursivePineapple: I changed it to `int store(ImmutableItemStack)`

src/main/java/com/gtnewhorizon/gtnhlib/item/InventoryItemSink.java
```diff
@@ -0,0 +1,121 @@
...
+            if (slot.matches(stack)) {
+                int slotLimit = getSlotStackLimit(iter.getCurrentSlot(), stack);
+                int remaining = slotLimit - slot.getStackSize();
+
+                int toTransfer = Math.min(remaining, stack.stackSize);
+
+                ItemStack rejected = iter.insert(stack.splitStack(toTransfer));
+
+                if (rejected != null) {
+                    stack.stackSize += rejected.stackSize;
```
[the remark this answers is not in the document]
RecursivePineapple: It gets decremented in `splitStack`, but I just removed that code entirely

RecursivePineapple:
Here's the summary of my most recent changes.

I pulled the getStack/setStack code out of `AbstractInventoryIterator` because that required implementations to have something like a slot. I've already made several iterators where this was not the case, so I just moved that code into `StandardInventoryIterator` and left `extract`/`insert` abstract. I also moved the `markDirty` code into `StandardInventoryIterator`.

I made all storing-related methods take an `ImmutableItemStack` and return an int which represents the number of rejected items. This will make the parameter properly immutable and reduce allocations, along with making everything simpler since you can't return an arbitrary ItemStack (that may not match what was given to the method). I added `InsertionItemStack` to make this simpler, which proxies a normal `ItemStack` or another `ImmutableItemStack`. It has a mutable stack size so that you can change it as you insert or extract items.

Other minor changes:
- Pulled the slot-detection logic for InventoryItemSource/InventoryItemSink into a protected method so that it can be overridden
- Removed the pointless stack size checks in InventoryItemSink
- Added a parameter called `forced` to `InventoryIterator.extract` and `InventoryIterator.insert` to allow `ItemTransfer.transfer` to inject rejected stacks easier
- Changed the default `test` in `ItemStackPredicate` to the `ItemStack` override instead of the `ImmutableItemStack` variant because this represents the intention better and prevents accidental mutation
- Add some useful factory methods to `ItemStackPredicate`
- Add `ItemStackPredicate.getStacks`, which will be helpful for AE interop
- Add a hash strategy for `ItemStack`, along with a generic one that supports `ItemStack`, `ItemId`, `ImmutableItemStack`. I implemented `ImmutableItemStack` on `ItemId` for this.

DarkShadow44:
Looks good to me know, thanks for sticking with me! Although it's still a bunch of complex code... Is this already tested and functional as far as you can see?

RecursivePineapple:
Yeah, I've been using this in three different mods. I have a few GT PRs that require it, along with an AE compat PR and a (not yet published) new Avaritia machine that requires it.

RecursivePineapple:
I've been testing my latest few changes in AE + Avaritia, though I haven't updated GT to use everything yet. It should be fairly simple since it just includes standard inventories.

RecursivePineapple:
I think you've been thorough enough :rofl:

PR #168, opened 2025-09-18
Fix memory leaks when creating shaders

cazmusw:
Shaders are already stored in the LinkProgram, it makes no sense to save them additionally.


kurrycat2004:
can you do a spotless apply

PR #176, opened 2025-10-22
BlockState System

RecursivePineapple:
This ports the block property system from MM and polishes up its rough edges. There's a proper BlockState class now, and a lot of the edge cases from MM's system have been cleaned up.

I haven't looked at the modern code for BlockStates because I wanted to keep this somewhat of a cleanroom implementation, so the API is probably completely different from how it looks in modern.

I know that modern has a huge problem with allocations, so there's a built-in pooling mechanism for BlockStates and they avoid as many allocations as possible. The map operations in BlockPropertyRegistry should still be the bottleneck, I can't profile this because my linux kernel still panics whenever I run spark :rofl:.


Nikolay-Sitnikov:
To be clear, with this explanation I have no objections to this idea. I was just confused by the comparison to modern BlockStates, since those function as a replacement for meta values with all the benefits of directly indexing combinations & with all the expected consequences when you have too many combinations of properties.

This is just an interface for sharing & editing common properties of blocks across mods, which a completely good idea.

RecursivePineapple:
@Nikolay-Sitnikov Interesting, that's a much more elegant system than we have in 1.7.10. That certainly sounds like it eats a lot of RAM though.

Nikolay-Sitnikov:
@RecursivePineapple As far as I know, that's about right. I do think that the modern system could be made more efficient/practical by applying a bit of dynamic code generation to the technique instead of using hashmaps for all the properties.

For instance, every class could declare its own immutable block-state class (inheriting from some parent `BlockState` class) to store the block-state information. Commonly used properties could be implemented with interfaces or possibly some kind of annotation processing, and the block states could be enumerated by reflectively reading annotations on the fields.

Something like:
```java
public class ObserverBlockState extends BlockState<BlockObserver>  implements IHasFacing<ObserverBlockState> {
    public final Direction facing;
    public final boolean active;

    // Automatically generated constructor - reflectively used to instantiate the variants.
    public ObserverBlockState(Block block, Direction facing, boolean active) {
        super(block);
        this.facing = facing;
        this.active = active;
    }

    // Automatically generated getter
    @Override
    public Direction getFacing() {
        return this.facing;
    }

    // Automatically generated transformer
    @Override
    public ObserverBlockState withFacing(Direction facing) {
        return BlockState.globalStates[/* implementation defined logic to find the index of the new state */];
    }
}
```

Getting a brand-new state would be a bit of a pain, though, since I can't figure out how to cleanly declare a method at code-writing-time with the signature dependent on the fields and the implementation being autogenerated.

RecursivePineapple:
@Nikolay-Sitnikov You'd have to make an annotation processor similar to lombok that generates those methods.

Annotation processors are given what is supposed to be a read-only AST, but you can actually modify it. I don't know why java hasn't made them properly mutable, but it's the only reason lombok works so the java devs aren't likely to change it.

I've tried to make my own (because I miss rust's AST macros) but it's pretty difficult to do so I haven't gotten very far.

Nikolay-Sitnikov:
@RecursivePineapple The problem is that I would also need to integrate it somehow with IDEs. (Which is why I was considering that dummy method idea.) Either way, it's not particularly likely that we can properly implement my improved version of the BlockState system into 1.7.10 (because it would require compatibility rewrites of every mod) and modern Minecraft already has the less efficient hashmap BlockStates.

sisyphussy (approved):
:shipit:

PR #195, opened 2025-11-10
Add ability to dynamically control block sounds

RecursivePineapple:
This adds an interface that allows blocks to control their step and break sounds dynamically. I'm going to use this in gt5u to control the sounds an ore block makes. I didn't see a reason to add a config for these mixins since they're all optional `WrapOperation`s.

I mixin'd everything in vanilla that refers to `Block.blockSound` so this should work in 99% of cases. I'm sure something in a mod somewhere doesn't work with this system, but that shouldn't matter. A block's sound is `stone` by default, so in the worst case scenario the block will just play the wrong sound.


RecursivePineapple:
@Dream-Master Should be good now

PR #200, opened 2025-11-15
More item moving fixes

RecursivePineapple:
- **Fix >64 stacks being cut off**
- **Fix NPE**

fixes: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/22524


Spicierspace153 (approved):
lgtm!

PR #208, opened 2025-11-19
Implement UV locking for JSON models

DarkShadow44:
Tested against an oak plank stair model, replicates the behavior from Minecraft. As examples the images from the Minecraft wiki [without](https://minecraft.wiki/w/File:Uvlock_false.png) and [with](https://minecraft.wiki/w/File:Uvlock_true.png) uv locking.

In theory my approach is fairly simple: When a model is rotated, we rotate the UVs in the other direction to keep the original orientation.


DarkShadow44:
I also added NEI and improved placement, just tell me if you want me to undo that.

PR #211, opened 2025-11-25
Add basic itemblock JSON models renderer

Shigure-Ruiseki:
Add basic item renderer and BlockColor from modern minecraft to use tintIndex

image

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/client/model/color/BlockColor.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/client/model/color/BlockColor.java
@@ -70,18 +70,18 @@ public class BlockColor {
 
         int color = -1;
 
-    // 1. Check for registered handler first, fallback to block's own IBlockColor implementation
-    IBlockColor handler = HANDLERS.get(block);
-    if (handler == null && block instanceof IBlockColor) {
-        handler = (IBlockColor) block;
-    }
-
-    if (handler != null) {
-        color = applyHandler(handler, world, stack, x, y, z, tintIndex);
-        if (color != -1) {
-            return color;
+        // 1. Check for registered handler first, fallback to block's own IBlockColor implementation
+        IBlockColor handler = HANDLERS.get(block);
+        if (handler == null && block instanceof IBlockColor) {
+            handler = (IBlockColor) block;
         }
-    }
+
+        if (handler != null) {
+            color = applyHandler(handler, world, stack, x, y, z, tintIndex);
+            if (color != -1) {
+                return color;
+            }
+        }
 
         // 3. ItemStack color
         if (stack != null && stack.getItem() != null) {


src/main/java/com/gtnewhorizon/gtnhlib/client/model/ModelISBRH.java
```diff
@@ -175,4 +215,80 @@ public static float diffuseLight(int normal) {
...
+                final float shade = diffuseLight(quad.getComputedFaceNormal());
+                tesselator.setColorOpaque_F(r * shade, g * shade, b * shade);
+                renderQuad(quad, -0.5f, -0.5f, -0.5f, tesselator, null);
+            }
+        }
+
+        // Rotated to exact side
+        GL11.glRotated(-90f, 0f, 1f, 0f);
+
+        // TODO add display
```
MalTeeez: If this means the JSON display property, could you specify that here. Its a little vague otherwise

src/main/java/com/gtnewhorizon/gtnhlib/client/model/color/IBlockColor.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Returns the color multiplier for an item stack.
+     * <p>
+     * Returning -1 means "do not override"; vanilla item color will be used.
+     *
+     * @param stack     The ItemStack instance.
+     * @param tintIndex The tint index, usually 0 for the main texture.
+     * @return Color multiplier as 0xRRGGBB or -1 to use default.
+     */
+    default int colorMultiplier(ItemStack stack, int tintIndex) {
```
MalTeeez: Same as L24

src/main/java/com/gtnewhorizon/gtnhlib/client/model/ModelISBRH.java
MalTeeez: My bad, I meant the TODO above that line

MalTeeez: Or actually right at L274 - not sure why github shows a different section

MalTeeez (changes requested):

src/main/java/com/gtnewhorizon/gtnhlib/client/model/ModelISBRH.java
Shigure-Ruiseki: Ok

src/main/java/com/gtnewhorizon/gtnhlib/client/model/color/BlockColor.java
```diff
@@ -0,0 +1,120 @@
...
+
+        if (color != -1) {
+            return color;
+        }
+
+        // 2. Registered handler
+        color = applyHandler(HANDLERS.get(block), world, stack, x, y, z, tintIndex);
+        if (color != -1) {
+            return color;
+        }
```
MalTeeez: These currently check twice for blocks that implement IBlockColor and are registered in HANDLERS.
I would combine them, so that its clear that registered handlers override the implementations.
Example:
```suggestion
    // 1. Check for registered handler first, fallback to block's own IBlockColor implementation
    IBlockColor handler = HANDLERS.get(block);
    if (handler == null && block instanceof IBlockColor) {
        handler = (IBlockColor) block;
    }

    if (handler != null) {
        color = applyHandler(handler, world, stack, x, y, z, tintIndex);
        if (color != -1) {
            return color;
        }
    }
```

MalTeeez (approved):

PR #223, opened 2025-12-06
Rewrite Universium Shader; Add Texture Classes

sisyphussy:
- Add classes for textures (TextureAtlas & AnimatedTexture). Those allow you to use animated textures outside of the TextureMaps from minecraft
- Rewrite & Relocate Universium Shader:
    - Use a static Texture Atlas for the texture (way more efficient, both on the CPU and GPU)
    - Cleanup the Shader (There was a bunch of useless junk in them)
    - Use uniforms to allow for customizations
    - Makes the color animation & star movement smoother
- Split TonemapShader from CustomFramebuffer
- Fix AutoShaderUpdater
- Remove redundant unbindFramebuffer() call in CustomFramebuffer creation

Currently, no code uses this class, so there won't be any issues, but I am planning on replacing avaritia's shaders with this (as it's more performant).

Tested extensively inside of GT5U


chrombread (approved):
Merging does no harm and lets it be tested in various places much more easily.

PR #228, opened 2025-12-09
Move villager names from WAWLA

SuperSoupr:
When mods register new villagers they aren't required to give forge any display name for them. WAWLA has a pretty huge collection of display names for villagers and is the only place in NH with them and it's already relied on by [mobs info here](https://github.com/kuba6000/Mobs-Info/blob/da8e173c1b4486096e637a8e7e91ab3e25e4dcd3/src/main/java/com/kuba6000/mobsinfo/nei/VillagerTradesHandler.java#L514).
We want to show the villager's profession's display name in UIE's version of the [trading post](https://github.com/GTNewHorizons/UtilitiesInExcess/pull/80), so moving this here is better than adding wawla as a dependency.
WAWLA already depends on GTNHLib through waila which depends on NEI which depends on GTNHLib.

Twin-PR https://github.com/GTNewHorizons/WAWLA/pull/8


chrombread (approved):
yay

PR #230, opened 2025-12-10
Add a simulated item sink interator

hiroscho:
Insertions on the simulated sink are not passed to the backing inventory

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/ClientProxy.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/ClientProxy.java
@@ -15,6 +15,7 @@ import net.minecraftforge.common.MinecraftForge;
 
 import com.gtnewhorizon.gtnhlib.client.model.ModelISBRH;
 import com.gtnewhorizon.gtnhlib.client.model.loading.ModelRegistry;
+import com.gtnewhorizon.gtnhlib.client.renderer.postprocessing.shaders.UniversiumShader;
 import com.gtnewhorizon.gtnhlib.client.tooltip.LoreHandler;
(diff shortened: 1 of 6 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/item/InventoryItemSink.java
```diff
@@ -103,4 +103,46 @@ protected int getSlotStackLimit(int slot, ItemStack stack) {
...
+            }
+
+            @Override
+            public ImmutableItemStack previous() {
+                throw new UnsupportedOperationException();
+            }
+
+            @Override
+            public boolean rewind() {
+                throw new UnsupportedOperationException();
```
RecursivePineapple: This should return false if rewinding isn't supported.

src/main/java/com/gtnewhorizon/gtnhlib/item/InventoryItemSink.java
```diff
@@ -103,4 +103,46 @@ protected int getSlotStackLimit(int slot, ItemStack stack) {
...
+            }
+
+            @Override
+            public ItemStack extract(int amount, boolean forced) {
+                throw new UnsupportedOperationException();
+            }
+
+            @Override
+            public ImmutableItemStack previous() {
+                throw new UnsupportedOperationException();
```
RecursivePineapple: Can you change this to a `return null` + a warn log? I can see this causing crashes in the future.

RecursivePineapple (commented):
Looks good, I have two minor comments.

RecursivePineapple (approved):

PR #236, opened 2025-12-17
Misc changes, and a few fixes

Roadhog360:
Here's what inside:
- MixinBlock_Particle.java:
  - Rename to MixinBlock_Particle.java (before the underscore wasn't there) to match how other mixins are named (also changes Mixin.java and GTNHLibConfig.java)
  - Particle icon wrapper now uses its own config option
  - Particle icon wrapper no longer checks if the block has an icon already defined (why do we care, JSON model icon should always take priority)
  - Particle icon wrapper now sets blockIcon to the missing texture if it's null, specifically for JSON models
    - This needs to be improved; see TODO
- JSONModel.java
  - Correct lightFace to be the cube's face name
  - Correct cullFace to be what sided store the quad is pooled into, instead of its normal, as well as now reading the `cullFace` field
    - Omni (ah-OOG-ah) has confirmed with me this is how it was supposed to be, and this way we can only run culling logic once and cull all faces in that pool with just one check
    - Quads without a cullface value set are now a part of the `UNASSIGNED` pool
  - Now sets the `shade`, `light_emission`, and `sprite` fields
  - Some fields have been elevated to protected for access within MissingModel.java
- MissingModel.java
  - Has been reworked to be a missing texture cube instead of a blank model
- ModelDeserializer.java
  - Now deserializes `lightEmission` field from JSON data
- ModelQuad.java, ModelQuadView.java, ModelQuadViewMutable.java
  - Add missing dirShading field - Is `shade` in the JSON, but is referred to as directionalShading and dirShading for the quad view funcs and field name, respectively for clarity
  - Add getters and setters that were previously missing
- ModelQuadFacing.java
  - Add functions to convert EnumFacing and ForgeDirection to ModelQuadFacing
- ModelISBRH.java
  - Now has behavior for `hasDirectionalShading`, culling faces, and emissiveness
    - Culling behavior is ignored while `RenderBlocks#renderAllFaces` is `true`
  - Bugfix: Quads in the `UNASSIGNED` pool are now rendered

TODO: (completed)
- ~~Culling is currently not rotated by the affine matrix, so blockstate files that define a rotation will have incorrect culling (this alone is why it's a draft PR, if this was fixed but got merged before the other TODOs were complete I'd make a new PR)~~
- ~Javadocs~

I've decided to move the remaining things that need to be done to a separate issue to comprehensively list all missing things about model JSONs. [You can view it here](https://github.com/GTNewHorizons/GTNHLib/issues/243)

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/cel/model/quad/properties/ModelQuadFacing.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/cel/model/quad/properties/ModelQuadFacing.java
@@ -20,7 +20,7 @@ public enum ModelQuadFacing {
     UNASSIGNED(0, 0, 0, null);
 
     public static final ModelQuadFacing[] VALUES = ModelQuadFacing.values();
-    public static final ModelQuadFacing[] DIRECTIONS = Arrays.stream(VALUES).filter(facing -> facing != UNASSIGNED)
+    public static final ModelQuadFacing[] DIRECTIONS = Arrays.stream(VALUES).filter(ModelQuadFacing::isDirection)
             .toArray(ModelQuadFacing[]::new);
     public static final ModelQuadFacing[] HORIZONTAL_DIRECTIONS = Arrays.stream(DIRECTIONS)
             .filter(facing -> facing.axis != Axis.Y).toArray(ModelQuadFacing[]::new);


Roadhog360:
Should be noted that even though I have stuff on the to-do, most of the code that is already written will likely be final aside for where a TODO is placed. So I'm already willing to take reviewers as I'm writing the code because I don't see much of this getting many revisions unless someone deems it necessary.

Roadhog360:
Due to low energy and time, I've decided I will not be completing my todo list for now. I'll open a separate PR when I'm ready to contribute again. For now I'll open an issue going over everything the system is missing and add to it at a later time. I apologize for setting everyone up for disappointment. I definitely will be coming back to this, but now is just not the time, IRL.

I'm leaving the cull face behavior in and just dumping everything in the UNASSIGNED list, so that way the only thing that needs to be done is just properly placing quads into the correct list.

cubefury:
all the best!

Roadhog360:
I've decided to move the remaining things that need to be done to a separate issue to comprehensively list all missing things about model JSONs. [You can view it here](https://github.com/GTNewHorizons/GTNHLib/issues/243)

This is to hopefully help other contributors and give me something to refresh myself with when I come back to this.

src/main/java/com/gtnewhorizon/gtnhlib/client/model/unbaked/JSONModel.java
```diff
@@ -214,13 +241,15 @@ public BakedModel bake(BakeData data) {
     }
 
     // TODO fix
-    private void bakeSprite(ModelQuadViewMutable quad, String name) {
+    // ... Fix what?
+    protected void bakeSprite(ModelQuadViewMutable quad, String name) {
```
ah-OOG-ah: That comment originally referred to UV mapping and rotation not working properly (it was way too early in the morning). If the icon actually gets set properly for all rotations of the model, please remove it.

Roadhog360: Rotation via blockstates woks, but UV rotation currently does not, as documented in my list. I can update the comment to reflect this.

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/cel/model/quad/ModelQuad.java
```diff
@@ -217,8 +239,6 @@ public void setState(int[] rawBuffer, int srcOffset, Flags flags, int drawMode,
         offsetPos(2, offsetZ);
 
         if (drawMode == GL_TRIANGLES) quadrangulate();
-
-        setLightFace(ModelQuadUtil.findLightFace(getComputedFaceNormal()));
```
ah-OOG-ah: Why was this line removed? This exists to set the light face for quads pulled out of the tesselator, not JSON models. Without it, the light face would be whatever the pooled quad had it set to.

src/main/java/com/gtnewhorizon/gtnhlib/client/model/ModelISBRH.java
```diff
@@ -75,14 +75,18 @@ public boolean renderWorldBlock(IBlockAccess world, int x, int y, int z, Block b
...
-                int stepZ = z + dir.getStepZ();
-                Block cullCheckBlock = world.getBlock(stepX, stepY, stepZ);
-                if (cullCheckBlock.isNormalCube(world, stepX, stepY, stepZ) && cullCheckBlock.isOpaqueCube()) {
-                    continue;
-                }
+            // Faces in the list that are not unassigned (isDirection() == true) check block.shouldSideBeRendered for
+            // that direction and cull that list if it's false. Uses a mixin to remove the bounds checks because those
+            // bounds have no effect on JSON geometry.
+            if (dir.isDirection() && !renderer.renderAllFaces
+                    && shouldSideBeRendered(
```
ah-OOG-ah: isn't this skipping if the side
- IS a direction (ok)
- and the renderer ISN'T rendering all faces (ok)
- and the side IS rendered (???)
I think this is missing a `!`

PR #245, opened 2025-12-28
Make JVMDG api stubs available for other mods

mitchej123:
Part of a larger set of work and forthcoming PRs to allow usage of JVMDowngrader for modern java compatability. This provides the stubs needed so mods don't have to shade it themselves.

Currently expected to be published as 0.9.0.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -4,13 +4,12 @@ configurations {
     jvmdgDowngraded17
 }
 
-def jvmDgVersion = "1.3.5-SNAPSHOT"
+def jvmDgVersion = "1.3.5"
 
 dependencies {
     shadowImplementation("it.unimi.dsi:fastutil:8.5.15") // Apache 2.0
     shadowImplementation("org.joml:joml:1.10.8") { transitive = false } // MIT
     shadowImplementation("com.mojang:brigadier:1.0.18") // MIT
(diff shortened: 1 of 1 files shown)


dependencies.gradle
```diff
@@ -7,6 +7,8 @@ dependencies {
     shadowImplementation("it.unimi.dsi:fastutil:8.5.15") // Apache 2.0
     shadowImplementation("org.joml:joml:1.10.8") { transitive = false } // MIT
     shadowImplementation("com.mojang:brigadier:1.0.18") // MIT
+    // 1.3.5-SNAPSHOT has some Buffer.slice() stubs that Angelica needs.  Switch to 1.3.5+ when available
+    shadowImplementation("xyz.wagyourtail.jvmdowngrader:jvmdowngrader-java-api:1.3.5-SNAPSHOT:downgraded-8") { transitive = false } // LGPL 2.1
```
eigenraven: Ideally we want to pull in at least downgraded-8 and downgraded-17 (not into shadowImplementation, probably in a new configuration in build.gradle.kts) and manually use some from() calls in the jar task in gradle to assemble a multiversion shaded copy of these so that on j17+ the j17 stubs are used for efficiency, I imagine that has the most efficiency gains and 17-25 will be much less

PR #246, opened 2025-12-29
Fix block cracks on JSON models

ah-OOG-ah:
The current strategy maps aligned vertex positions directly to UVs on the block crack texture - this mostly matches modern, bar some incorrect rotations. Unaligned quads instead map their original UVs to UVs on the block crack texture, which works well for *normal* quads but may break down for modded ones. I'm unsure what modern does here yet.
- [ ] fix rotation on aligned quads
- [ ] copy modern behavior on unaligned quads

Also adds localization for test blocks and a lectern model.


Roadhog360:
Tbh I think it's practically impossible to fix this without some mixin to `RenderGlobal#drawBlockDamageTexture`.

Roadhog360:
Spotless moment

ah-OOG-ah:
superceded by #343

PR #257, opened 2026-01-12
Document known issue with Fnv1a algorithms in javadoc

WNP78:
Following discussion on discord, adding known issue to inline documentation. Electing not to fix it as it would break existing worldgen seeds.

Explanation of the bug:
The Fnv1a algorithm hashes a single octet (byte) of input in a step, so any input is split into individual bytes to be processed, where that byte is XOR'ed with the lower 8-bits of the state, which is then multiplied by a large prime.

In the java implementation, each octet is stored as a `byte` value which is treated as signed by Java. When it is XOR'ed with the state, both operands must be the same size, so the byte, being smaller than the state (32 or 64 bits), is expanded. Because Java treats it as signed, this means that if the sign (upper) bit of the hashed byte is set, then when it's expanded to 32 or 64 bits the added bits will all be set to 1 to preserve the integer value. This means that the upper bits of the state are flipped. Since all the bits are flipped, it should not bias the value in any significant manner, though I have done no formal study of the mathematics behind the hash algorithm.


wlhlm (approved):
Thank you!

Issue #265, opened 2026-02-06
Number formatting doesn't respect Minecraft's language settings


Eldrinn-Elantey:
# Problem Description

Currently, `NumberFormatUtil` uses `Locale.getDefault(Locale.Category.FORMAT)` which retrieves the system's JVM locale, not Minecraft's selected language. This means that when a player changes the game language in Minecraft settings, number formatting (thousand separators, decimal points) doesn't change accordingly.

**Example:**
- Player launches client with system locale `en_US`
- Numbers display as: `1,024` (comma separator)
- Player switches Minecraft language to Russian (`ru_RU`)
- Numbers still display as: `1,024` instead of expected `1 024` (space separator)

This affects all tooltips, GUIs, and any place where `NumberFormatUtil.formatNumber()` is used (pumps, machines, energy values, etc.).

# Attempted Solution

I tried to fix this by:
1. Adding a method `getMinecraftLocale()` that reads the current language from `Minecraft.getMinecraft().getLanguageManager().getCurrentLanguage()`
2. Caching the locale on first use to avoid repeated lookups
3. Using this locale in `getDefaultDecimalFormatter()` instead of system locale

**Code changes:**
```java
@SideOnly(Side.CLIENT)
private static Locale getMinecraftLocale() {
    if (cachedLocale == null) {
        try {
            String langCode = Minecraft.getMinecraft()
                .getLanguageManager()
                .getCurrentLanguage()
                .getLanguageCode();
            
            String[] parts = langCode.split("_");
            if (parts.length == 2) {
                cachedLocale = new Locale(parts[0], parts[1]);
            } else {
                cachedLocale = new Locale(parts[0]);
            }
        } catch (Exception e) {
            cachedLocale = Locale.getDefault(Locale.Category.FORMAT);
        }
    }
    return cachedLocale;
}

private static Locale getLocale() {
    if (FMLCommonHandler.instance().getSide().isServer()) {
        return Locale.getDefault(Locale.Category.FORMAT);
    }
    
    try {
        return getMinecraftLocale();
    } catch (Exception e) {
        return Locale.getDefault(Locale.Category.FORMAT);
    }
}
```

# Result

**Initial attempt with per-tick update caused screen flickering.**

# Questions

1. Is this the correct approach, or is there a better way to get Minecraft's current locale?
2. Should we listen to a language change event to update the cache dynamically, or is restart-required acceptable?
3. Any concerns about performance or compatibility with this implementation?

PR #266, opened 2026-02-08
Fix/use game locale for numbers

Eldrinn-Elantey:
feat: Add locale selection for number formatting with live reload
Implements a configurable locale override system for number formatting
to allow
users to customize how numbers are displayed regardless of their system
locale.

Features:
- Dropdown menu with 17 predefined locales (en_US, ru_RU, de_DE, fr_FR,
etc.)
- Each option shows formatting example (e.g., "Russian - 1 234,56")
- System Default option to use OS locale settings
- Live reload - changes apply immediately without game restart
- Config event handler to sync settings when modified in GUI

Technical changes:
- Added LocaleOption enum to NumberFormatConfig
- Added ConfigEventHandler for live config reloading
- Modified getDefaultDecimalFormatter() to use configurable locale
- Added syncNumberFormatting() initialization in CommonProxy
- All locale parsing and error handling centralized in NumberFormatConfig

java_jW3BfudUAp

The solution to this problem https://github.com/GTNewHorizons/GTNHLib/issues/265


src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatUtil.java
```diff
@@ -229,8 +236,44 @@ private static boolean isExponentialFormattingEnabled(BigDecimal absValue, Numbe
...
+            FMLCommonHandler handler = FMLCommonHandler.instance();
+            if (handler != null && handler.getSide().isServer()) {
+                return Locale.getDefault(Locale.Category.FORMAT);
+            }
+        } catch (Throwable ignored) {
+            return Locale.getDefault(Locale.Category.FORMAT);
+        }
+
+        try {
+            return getMinecraftLocale();
```
Nikolay-Sitnikov: This cache does not get cleared when the player changes the language.

Nikolay-Sitnikov (dismissed):
Instead of caching the locale when the call is made, I suggest making the locale static field default to `Locale.getDefault(Locale.Category.FORMAT)`, while using `((IReloadableResourceManager) Minecraft.getMinecraft().getResourceManager()).registerReloadListener` in the client-side `preInit` event handler in order to bind a method that would overwrite the default when the player loads their language.

Also, a config setting to force a specific formatting locale would be nice.

src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java
```diff
@@ -26,6 +29,126 @@ public final class NumberFormatConfig {
...
+                    "Number formatting: Using system default locale ("
+                            + Locale.getDefault(Locale.Category.FORMAT).toString()
+                            + ")");
+        } else {
+            GTNHLib.info(
+                    "Number formatting: Using " + numberFormatLocale.getDisplayName()
+                            + " ("
+                            + customLocale.toString()
+                            + ")");
+        }
```
Nikolay-Sitnikov: (This would also need to be rewritten to care about the locale - you could probably simplify it to just a single logging statement, though, or check against the selected enum.)

Dream-Master:
@Eldrinn-Elantey your pr throw a few errors when run testes and need to be fixed
image

leagris (dismissed):

leagris (dismissed):

leagris (dismissed):

leagris (dismissed):

leagris (dismissed):

leagris (commented):
.

leagris (changes requested):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java#L58
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/eventhandlers/ConfigEventHandler.java#L6

`ConfigChangedEvent` is from `cpw.mods.fml.client.event`, a client-side package. Registering this handler in `CommonProxy` will crash on a dedicated server with `ClassNotFoundException`. Move the registration to `ClientProxy`.

leagris (changes requested):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/ExponentialFormat.java#L57
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/ExponentialFormat.java#L102

Both `formatScientificStandard` and `formatScientificPowerOfTen` still call `Locale.getDefault(Locale.Category.FORMAT)` directly, bypassing `NumberFormatConfig.getActiveLocale()`. The locale override has no effect on exponential values, which is the core of what this PR is supposed to fix.

leagris (changes requested):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java#L112-L117

The catch is broader than needed since `ExponentialFormat.parse` only throws `IllegalStateException`. Narrow it:
```java
} catch (IllegalStateException ignored) {
    GTNHLib.error("Invalid exponential format pattern: '" + formatPattern + "', using SCIENTIFIC");
    EXPONENTIAL_FORMAT = ExponentialFormat.SCIENTIFIC;
}
```

leagris (changes requested):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java#L57
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java#L132-L137

`customLocale` can be stale before `syncNumberFormatting()` is first called, silently returning the wrong locale. `LocaleOption.getLocale()` already handles `SYSTEM_DEFAULT`, so simplify to:
```java
public static Locale getActiveLocale() {
    return numberFormatLocale.getLocale();
}
```

leagris (commented):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java#L64

`SYSTEM_DEFAULT` captures the system locale at class-load time into the `locale` field, which is dead state since `getLocale()` overrides it. Pass `null` instead and null-check in `getLocale()`.

leagris (commented):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/eventhandlers/ConfigEventHandler.java#L15-L16
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java#L120-L124

Two INFO lines emitted for a single config change. Also, "changes applied immediately" logs unconditionally even when the fallback to `SCIENTIFIC` was triggered. Remove the log from `ConfigEventHandler`, keep only the one in `syncNumberFormatting` with a distinct message for the fallback path.

leagris (commented):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java#L144-L146

`hasCustomLocale()` is not called anywhere. Either remove it or add a Javadoc noting it is intended for downstream consumers.

leagris (commented):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/resources/assets/gtnhlib/lang/en_US.lang#L98

Tooltip says "Leave empty for system default" but the field is an enum dropdown. Suggested replacement: "Select the number formatting locale. Choose SYSTEM_DEFAULT to use your operating system locale."

leagris (commented):
https://github.com/GTNewHorizons/GTNHLib/blob/bde930af4de9434fd9d74f602974ba17fe42f9ca/src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java#L24-L28

`@Config.Comment` for `formatPattern` says "Invalid values will cause the game to fail fast with an error." Actual behaviour is a logged error and silent fallback to `SCIENTIFIC`. Update to reflect the real behaviour.

src/main/java/com/gtnewhorizon/gtnhlib/util/numberformatting/NumberFormatConfig.java
```diff
@@ -22,10 +25,124 @@ public final class NumberFormatConfig {
...
+            fallbackToScientific = true;
+        }
+
+        Locale activeLocale = numberFormatLocale.getLocale();
+        if (fallbackToScientific) {
+            GTNHLib.info(
+                    "Number formatting: Using " + numberFormatLocale.getDisplayName()
+                            + " ("
+                            + activeLocale
+                            + "); invalid format pattern, fell back to SCIENTIFIC");
```
leagris: It is a bit weird that if the exponential format pattern is invalid it will log both an error message and an info message.
Not a blocker

leagris (approved):
Apart from curious non-blocking ERROR + INFO message combined, it is all good to go

Dream-Master:
@Nikolay-Sitnikov

Dream-Master:
@Eldrinn-Elantey

leagris (approved):
спасибо большое

PR #281, opened 2026-02-20
Add IConfigEntry to the Config system

0hwx:
Now you can add Custom Entries for fields or categories 

here is an example I was working on 

Screenshot from 2026-02-20 07-33-06

Screenshot from 2026-02-20 07-33-43

image

image


slprime:
It would be great if someone added field sorting annotation (in another PR)

Caedis:
Can you add an example config that uses this?
Register it only in dev envs

Caedis (approved):

0hwx:
> It would be great if someone added field sorting annotation (in another PR)

You can use the entries to sort the configs like this:
image

Caedis:
neat

PR #282, opened 2026-02-20
Add InventoryChangedEvent

Caedis:
Uses a centralized player inventory scanner to replace the multiple ones in the pack like BQ and GC/GS

Monitors the player's cursor, inventory, armor, 2x2 craft matrix, and baubles (if baubles is loaded).
Tracks via  item id and meta (no NBT) using Long2IntOpenHashMap

Added `InventoryChangedEvent.ItemAdded` and `InventoryChangedEvent.ItemRemoved` events that posts to the client and server side when the player's inventory changes.

Does not trigger when an item changes slots (sorting), only if it truly enters or leaves their inventory.

Added a config option to throttle the scanner `inventoryScanInterval`. Default is every 5 ticks.

Only runs when something is subscribed to any of its events.

Preview with debug handler
The crafting table craft was using the player's 2x2 crafter (both showed up upon picking up the output)
The door craft was using a normal crafting table
image

Will be very useful for AmazingTrophies so they are not tied to the pickup event

AI Transparency: claude was used in the making of this pr


src/main/java/com/gtnewhorizon/gtnhlib/event/inventory/InventoryChangeScanner.java
```diff
@@ -0,0 +1,176 @@
...
+public final class InventoryChangeScanner {
+
+    private static final Object2ObjectMap<UUID, PlayerScanState> SERVER_STATES = new Object2ObjectOpenHashMap<>();
+    private static final Object2ObjectMap<UUID, PlayerScanState> CLIENT_STATES = new Object2ObjectOpenHashMap<>();
+
+    private InventoryChangeScanner() {}
+
+    @SubscribeEvent
+    public static void onPlayerTick(TickEvent.PlayerTickEvent event) {
+        if (!GTNHLibConfig.inventoryScannerEnabled) return;
```
boubou19: Is it reasonable to make it a config? If False, BQ and other mods depending on it will never get the events they need to properly work. While i get it that if you don't use such a mod, the ability to toggle this off for perf gains seems interesting, the debugging can be a little tricky in the case the config is off but then an user adds a mod using that feature. One way around this would be to make sure every mod using this event would add a check against this config and tell the player it needs it to function properly, but that seem brittle. Have you done some benchmark yet? Because this PR does not do ASM/mixins, it should be safe to leave it on imo.

Caedis: Maybe default off, then a require mods to set it to true if it needs it?
The main performance impact i could see is the NBT hashing

Working on getting a setup created to profile this in the full pack

Caedis:
Hmm, might do what BQ does and attach to the player's inventoryContainer instead of always scanning

Caedis:
@Connor-Colenso was the ae2 issue on the client side or the server side?

Caedis:
Ready for review

Tested with AE2 in a devenv and was not able to recreate @Connor-Colenso's issue

Testing in full pack now

Caedis:
Seems to be good in the full pack. Tested AE2 as well

Algent (approved):
let's go, now you have to make BQ and others use it 🤣

PR #284, opened 2026-02-22
Changes need to support Core Profile in Angelica

mitchej123:


src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/vao/VertexBufferStorage.java
```diff
@@ -7,23 +7,30 @@
...
 
+    private static final VaoFunctions vao = GLCaps.VAO;
+
     private int id;
     private int vertexCount;
     private final VertexFormat format;
     private final int drawMode;
     private final int flags;
+    private int vaoId;
+    private boolean vaoDirty;
```
sisyphussy: Why's this needed? I'd like to keep vbo's and vao's separate, and I'm not too sure how well this will work for any IVertexArrayObject that encapsulates the buffer

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/vertex/VertexFormatElement.java
```diff
@@ -76,29 +77,48 @@ public enum Usage {
...
 
         private void setupBufferState(int size, int type, int stride, long pointer, int index) {
-            this.setupState.setupBufferState(size, type, stride, pointer, index);
+            if (GLCaps.isCoreProfile()) {
+                if (this == PADDING) return;
+                final int location = attributeLocation >= 0 ? attributeLocation : index;
+                GL20.glVertexAttribPointer(location, size, type, normalized, stride, pointer);
+                GL20.glEnableVertexAttribArray(location);
+            } else {
+                this.setupState.setupBufferState(size, type, stride, pointer, index);
```
sisyphussy: This will work, but it'll be limited to GTNHLib functionality (mods that call glEnableClientState etc will not work properly). Shouldn't this be implemented by AngelicaRedirector instead?

sisyphussy: Could be that i'm missing some context here since I don't think you've published the core changes to a branch yet

Ranzuu (approved):
They made me do it

PR #288, opened 2026-02-24
Prevent resizing the ClassConstantPoolParser during runtime

Alexdoru:
remove the capability to add strings to search to the ClassConstantPoolParser because it shouldn't be dynamically resized during runtime in order to avoid unforeseen performance issues


Alexdoru:
> https://github.com/GTNewHorizons/Angelica/blob/1d96af8e3aeef83026a9a142fba0027cfa3e948a/src/main/java/com/gtnewhorizons/angelica/loading/shared/transformers/AngelicaRedirector.java#L314
> 
> Seems like this'll need fixing (not sure if you're aware of it)

already fixed + released a new angelica build

PR #295, opened 2026-03-03
Implement blockstate to Json Model

Shigure-Ruiseki:
This pull request implements a BlockState system for the Json Model loader.

Instead of being limited to metadata for model selection, JsonModel can now read and resolve models using the blockstate system. This allows model variants to be selected based on full block state data, including values stored in TileEntity, rather than being restricted to meta values only.

As a result, blocks are no longer constrained by the 4-bit metadata limit and can support more complex and dynamic rendering behavior.

And make display can read from parent.

image

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinModelFHC.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinModelFHC.java
@@ -24,104 +24,106 @@ import net.minecraftforge.client.MinecraftForgeClient;
 import org.lwjgl.opengl.GL11;
 import org.lwjgl.opengl.GL14;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 
 import com.gtnewhorizon.gtnhlib.client.model.ModelISBRH;
-import com.llamalad7.mixinextras.injector.wrapmethod.WrapMethod;
-import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 
 @Mixin(value = ForgeHooksClient.class, remap = false)
 public class MixinModelFHC {
 
-    @WrapMethod(method = "renderInventoryItem")
-    private static boolean nhlib$renderModeledInventoryItem(RenderBlocks renderBlocks, TextureManager engine,
-            ItemStack item, boolean inColor, float zLevel, float x, float y, Operation<Boolean> original) {
+    @Inject(method = "renderInventoryItem", cancellable = true, at = @At("HEAD"))
+    private static void renderModeledInventoryItem(RenderBlocks renderBlocks, TextureManager engine, ItemStack item,
+            boolean inColor, float zLevel, float x, float y, CallbackInfoReturnable<Boolean> cir) {
         IItemRenderer customRenderer = MinecraftForgeClient.getItemRenderer(item, INVENTORY);
-        if (!(customRenderer instanceof ModelISBRH modelISBRH)) {
-            return original.call(renderBlocks, engine, item, inColor, zLevel, x, y);
-        }
-        engine.bindTexture(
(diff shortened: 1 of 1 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/client/model/loading/ModelDeserializer.java
```diff
@@ -96,12 +96,11 @@ private static Map<Position, ModelDisplay> loadDisplay(JsonObject in) {
 
                 final String name = j.getKey();
                 final Position pos = Position.getByName(name);
-                ret.put(pos, loadADisplay(j.getValue().getAsJsonObject()));
-            }
-        }
 
-        for (Position p : Position.values()) {
-            ret.putIfAbsent(p, ModelDisplay.DEFAULT);
```
ah-OOG-ah: Why was this line removed? It looks like no default positions are being loaded now.

MalTeeez (commented):
Looks good aside from the missing default positions and:
Could you document this feature (if only briefly) in the models.md, as it will probably be missed by people who dont see this PR otherwise.

Works: 
image

src/main/java/com/gtnewhorizon/gtnhlib/client/model/state/MonopartState.java
```diff
@@ -14,52 +17,68 @@
...
+                                + c
+                                + "'!");
 
-                states.put(c.substring(0, eidx), c.substring(eidx + 1));
+                states.put(c.substring(0, eqIndex), c.substring(eqIndex + 1));
             }
         }
 
-        boolean matches(Map<String, String> properties) {
+        public boolean matches(Map<String, String> properties) {
```
ah-OOG-ah: Why is this public now?

src/main/java/com/gtnewhorizon/gtnhlib/client/model/loading/ModelDeserializer.java
Shigure-Ruiseki: im make it to json model check and read display from parent if parent null too it will set default by PileOfQuads getDisplay

src/main/java/com/gtnewhorizon/gtnhlib/client/model/ModelISBRH.java
```diff
@@ -398,7 +417,115 @@ private void applyItemDisplay(BakedModel model, int meta, ItemRenderType type) {
...
+        worldContext.x = x;
+        worldContext.y = y;
+        worldContext.z = z;
+        worldContext.random = RAND;
+        worldContext.blockState = BlockPropertyRegistry.getBlockState(world, x, y, z);
+        final var model = getModel(worldContext);
+        return model.getParticle(worldContext);
+    }
+
+    private static class WorldContext implements BakedModelQuadContext.World {
```
ah-OOG-ah: It's probably better if this class (as well as the item one) is broken out into a separate file, not as an inner class. The internals are trivial and don't need to be inside the ModelISBRH class. Also, it should probably have protected or public access, since people are expected to be able to implement ModelISBRH themselves, if they wish.

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinEntityDiggingFX_IconWrapper.java
```diff
@@ -0,0 +1,33 @@
...
+
+@Mixin(EntityDiggingFX.class)
+public class MixinEntityDiggingFX_IconWrapper {
+
+    @WrapOperation(
+            method = "<init>(Lnet/minecraft/world/World;DDDDDDLnet/minecraft/block/Block;II)V",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/block/Block;getIcon(II)Lnet/minecraft/util/IIcon;"))
+    private IIcon nhlib$useWorldIcon(Block block, int side, int meta, Operation<IIcon> original, World world, double x,
+            double y, double z, double mx, double my, double mz, Block passedBlock, int passedMeta, int passedSide) {
+        if (block.getRenderType() == ModelISBRH.JSON_ISBRH_ID) {
```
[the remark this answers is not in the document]
Shigure-Ruiseki: i should make interface to check it?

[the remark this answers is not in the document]
Shigure-Ruiseki: I currently have no idea what changes might be made if I add an interface to customize the model like IModelProvider for the block.

PR #296, opened 2026-03-03
Fix EIr only passing the replacements once per world restart

LazyFleshWasTaken:
Fix the Enhanced infusion recipe only passing the replacement list to the mixin once, as its the actual list and not a copy, and so the clear at the end of the infusion matrix run is probably deleting it. 

~~Am too tired to test, will undraft when i do tomorrow.~~

Fix looks to have worked. Could not do more than one before a restart on this instance, with the fix I can. 
image


LazyFleshWasTaken:
Ok, uh, apperantly build check is failing, but that seems to be due to maven issues

Caedis (approved):

PR #297, opened 2026-03-04
GTNHTeams: Arbitrary-data team system

FourIsTheNumber:
I adapted this implementation from Maya's work on our April Fools 2025 pack. This is a teams system that can store arbitrary implementations of ITeamData and can therefore be used to centralize all of our many teams systems. GTNHTeams can be managed using brigadier commands by default.

Any data can be attached to a GTNHTeam by creating an implementation of ITeamData with at least read/write NBT methods, then registering it to the TeamDataRegistry. Each implementation of ITeamData will be associated with each team and saved to the world.

### Team System:

-Players are added to a personal team immediately on join.
-Teams are multi-owner. The creator is immediately made an owner.
-Any owner can promote members to officer, officers to owner, or demote vice versa.
-Only owners and officers can send invites to other players.
-Teams can be merged by sending a merge request to another team. Only owners can send, accept, or deny merge requests. After a merge, all members of the consumed team are added to the new team as members. mergeData() will be called on all ITeamData - what data merging looks like is up to the implementer.
-If a player accepts an invite to join a team from a team where they are the sole member, the team is disbanded and data is merged to the new team.
-A player cannot leave a team where they are the sole member, except through invite or merge to another team.
-Owners cannot demote themselves or leave the team if they would leave it in a state with no owners.

### Player-Side Commands:

Owners-Only:
/gtnhteam promote [player name]: Promote a member to officer or an officer to owner.
/gtnhteam demote [player name]: Demote an owner to officer or an officer to member.
/gtnhteam rename [team name]: Rename your team.
/gtnhteam merge request [team name]: Request for your team to be merged into another team, transferring its members and merging data.
/gtnhteam merge accept (opt:team name): Accept merge request. If team only has one pending request, they don't need to specify.
/gtnhteam merge deny (opt:team name): Deny merge request. If team only has one pending request, they don't need to specify.

Officers-Only:
/gtnhteam invite [player name]: invite a player to your team.

Anyone:
/gtnhteam: List commands.
/gtnhteam help: Print command usage details.
/gtnhteam leave: Exit a multiplayer team and automatically create a solo team.
/gtnhteam info: List team name and ranks of members.
/gtnhteam accept (opt: team name): Accept an invite. If player only has one pending invite, they don't need to specify.
/gtnhteam deny (opt: team name): Decline an invite. If player only has one pending invite, they don't need to specify.

### Operator Commands:

/gtnhteam_admin: List commands.
/gtnhteam_admin help: Print usage details.
/gtnhteam_admin rename [team name] [new name]: Forcibly rename a team.
/gtnhteam_admin promote [team name] [player name]: Forcibly promote a player one rank.
/gtnhteam_admin demote [team name] [player name]: Forcibly demote a player one rank.
/gtnhteam_admin merge [source team] [target team]: Forcibly merge source into target.
/gtnhteam_admin info [team name]: List ranks of members in target team.
/gtnhteam_admin disband: Forcibly disband a team, returning all players in it to solo teams and destroying all team data for the disbanded team.


MeiTianyou (approved):
looks good boss

src/main/java/com/gtnewhorizon/gtnhlib/teams/Team.java
```diff
@@ -0,0 +1,123 @@
...
+        return owners.contains(player);
+    }
+
+    public void addMember(UUID uuid) {
+        if (!members.contains(uuid)) members.add(uuid);
+        TeamWorldSavedData.markForSaving();
+    }
+
+    public void addOfficer(UUID uuid) {
+        if (!officers.contains(uuid)) officers.add(uuid);
```
Glease: same here.

src/main/java/com/gtnewhorizon/gtnhlib/teams/Team.java
```diff
@@ -0,0 +1,123 @@
...
+public class Team {
+
+    @Getter
+    private String teamName;
+    private final ObjectList<UUID> owners = new ObjectArrayList<>();
+    private final ObjectList<UUID> officers = new ObjectArrayList<>();
+    private final ObjectList<UUID> members = new ObjectArrayList<>();
+    private final Map<String, ITeamData> teamData = new HashMap<>();
+
+    public Team(String teamName) {
```
Glease: If client code is supposed to get Team instances solely from TeamManager and never construct a Team instance on their own, this should probably be package private.
If my assumption does not hold, there should be some documentation stating when and why that could happen.

Glease (commented):
there are two additional points I'd like to discuss
1. Most command send feeback only to the one running the command, but in many cases there are more than one people that are interested in the event happened. e.g. when someone joins, everyone (or at least the team owners) would want to know that the person in question has accepted our invite. This can be delayed for a later patch since this is really just an implementation detail, not a critcial API.
2. I'm actually not too fond of storing team in WorldSavedData as an NBT tag. it's not the most readable form of data structure, making server administration harder. The in game admin command does mostly remove the need to poke at the underlying storage file, but there are always cases of downtime when executing a command is not really possible. If you don't object I can cook up a PR to change to JSON

src/main/java/com/gtnewhorizon/gtnhlib/teams/Team.java
```diff
@@ -0,0 +1,123 @@
...
+    public boolean isTeamOfficer(UUID player) {
+        return officers.contains(player);
+    }
+
+    public boolean isTeamOwner(UUID player) {
+        return owners.contains(player);
+    }
+
+    public void addMember(UUID uuid) {
+        if (!members.contains(uuid)) members.add(uuid);
```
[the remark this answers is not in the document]
FourIsTheNumber: These are not sets in the current implementation - but they probably should be.

Luca-Guettinger:
I’m currently working on a genetics refactor, and with a change like this it would be pretty easy to let all team members share their genetics progress.

Right now, executeLeave() and executeAdminDisband() don’t copy any ITeamData to the player’s new solo team. That means any mod that stores shared data on the team (like progress, unlocks, discoveries, etc.) just disappears for that player when they leave.

Would it make sense to add a data copy step after creating the solo team? Maybe something like a copyForLeavingPlayer(UUID) default method on ITeamData that returns the data to copy over (or null if nothing should be copied). This could be called for each registered data key, similar to how mergeData works for merges.

There’s a similar issue in executeAccept(): if a player joins a new team while their old team still has members, the old team keeps its data, but the player’s new team starts fresh with no carry-over.

tiffit:
Made the following changes:
- Provided teams for some methods in `ITeamData`
- Added `copyData` method to ITeamData so implementations can copy data when a player moves teams
- Teams are given a UUID as the team name can change
- Teams are saved as json files in `saves/{world name}/gtnhteams/`
- Team UUID and name are synced to the client
- `ITeamData` implementations can instead implement `INetworkTeamData` if they want to sync data to the client. They must call `TeamNetwork#syncTeamData` to actually have the data sync. Data is also synced on player login and when a player joins the team.
- Created `TeamCreateEvent` and `TeamMergeEvent`

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamManager.java
```diff
@@ -0,0 +1,206 @@
...
+    public static void copyTeamData(Team oldTeam, Team newTeam, UUID playerId, TeamDataCopyReason reason) {
+        for (Entry<String, ITeamData> entry : oldTeam.getAllDataEntries()) {
+            ITeamData copied = entry.getValue().copyData(oldTeam, newTeam, playerId, reason);
+            if (copied != null) {
+                newTeam.putData(entry.getKey(), copied);
+            }
+        }
+    }
+
+    public static void ForEachOnlineTeamMember(Team team, Consumer<EntityPlayerMP> consumer) {
```
cubefury: minor lower camel case inconsistency

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamManagerClient.java
```diff
@@ -0,0 +1,52 @@
...
+public class TeamManagerClient {
+
+    private static Team TEAM;
+
+    @SubscribeEvent
+    private static void onDisconnect(ClientDisconnectionFromServerEvent event) {
+        TEAM = null;
+    }
+
+    public static void OnTeamInfoSyncPacket(TeamInfoSync packet) {
```
cubefury: here too

cubefury (changes requested):
tiny things

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamAdminCommand.java
```diff
@@ -0,0 +1,231 @@
...
+        TeamManager.forEachOnlineTeamMember(team, player -> NetworkHandler.instance.sendTo(packet, player));
+
+        return success(
+                sender,
+                "gtnhlib.chat.teams.admin.message.renamed",
+                colorChatComponent(EnumChatFormatting.GOLD, oldName),
+                colorChatComponent(EnumChatFormatting.GOLD, newName));
+    }
+
+    private static int executeAdminPromote(ICommandSender sender, String teamName, String playerName) {
```
cubefury, line 102: Do we want to check the sender's rank if it's an officer and block promotion/demotion above their rank as well? Or we can just leave it as server op discretion.

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamAdminCommand.java
```diff
@@ -0,0 +1,231 @@
...
+        TeamManager.mergeTeams(targetTeam, sourceTeam);
+
+        ChatComponentTranslation notification = new ChatComponentTranslation(
+                "gtnhlib.chat.teams.message.merge_complete",
+                colorChatComponent(EnumChatFormatting.GOLD, sourceName),
+                colorChatComponent(EnumChatFormatting.GOLD, targetName));
+        notification.getChatStyle().setColor(EnumChatFormatting.GREEN);
+
+        for (UUID memberUuid : allMembers) {
+            EntityPlayer member = sender.getEntityWorld().func_152378_a(memberUuid); // getPlayerByUUID
```
cubefury, line 162: In the future if we allow open to LAN I think this might need to change, since iirc LAN player UUIDs are a bit wonky. (This is why BQ and vending machine use a NameCache class, but I could just be running off outdated info). We can address this at some point in the future!

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamAdminCommand.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        return success(
+                sender,
+                "gtnhlib.chat.teams.admin.message.merged",
+                colorChatComponent(EnumChatFormatting.GOLD, sourceName),
+                colorChatComponent(EnumChatFormatting.GOLD, targetName));
+    }
+
+    private static int executeAdminDisband(ICommandSender sender, String teamName) {
```
cubefury, line 173: Also in a similar fashion here, I'm not sure if we want to allow players to disband a team they're not a part of

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamDataSaver.java
```diff
@@ -0,0 +1,220 @@
...
+                }
+            } catch (Exception ex) {
+                GTNHLib.LOG.error("Error while loading TeamData {} for team {}", key, uuid, ex);
+            }
+        }
+        return team;
+    }
+
+    private void saveToFiles() {
+        if (!saveDir.exists()) {
```
cubefury: let's makedirs here instead of aborting, and fail only if the os doesn't give us write permissions.

cubefury (changes requested):
1. I think we should disallow players to mutate teams they are not a part of, unless they're an op of the server or running the command through the server console. This is because team maangement wont be exclusive to server op.
2. The file save comment for makedirs instead of failing immediately

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamAdminCommand.java
tiffit: I think we should leave as much control as possible to the server OP

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamAdminCommand.java
tiffit: Same as above, this is an admin command and the goal is to give as much power to admins as possible via commands

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamDataSaver.java
tiffit: Done

cubefury (approved):
lgtm!

cubefury (approved):
ye

PR #312, opened 2026-03-14
Set colorIdx to -1 default

Ruling-0:
Every place that references `colorIdx` appears to check if it's -1 as some default. This includes:
- `ModelQuadView#hasColor` (returns true if colorIdx != -1).
- `ModelISBRH`, where if the index is != -1 it gets the color for a quad from an underlying BlockColor, and if it is == -1, just uses the model's single color.

So, this PR sets the default in `ModelQuad` to be -1.


ah-OOG-ah (approved):
Note that the color system is currently... not ok. The reason why the default color is -1 is because that translates to `0xFFFFFFFF`, or pure white. I have no idea what happened to colors while I wasn't looking, but it is *not* supposed to be this convoluted.

Anyways, seems reasonable.

PR #313, opened 2026-03-15
Add proper automatic model detection

ah-OOG-ah:
Refactor model detection (yet again, yipee!). Now, merely *having* a state->model JSON (in a resourcepack or registered mod, as usual) is sufficient to model a block. 

Overriding the ISBRH for custom model loading is no longer supported; instead, you need to implement `BlockModelInfo` and `IBlockModelProvider` on your block.

Documentation on the blockstate-enabled model system has been improved, and two helper methods added.

Finally, break particles now properly take the model's particle texture. And also I added lang entries for the test blocks.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/core/fml/transformers/BlockIconTransformer.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/core/fml/transformers/BlockIconTransformer.java
@@ -41,6 +41,7 @@ public class BlockIconTransformer implements IClassTransformer {
     private static final String NEW_WORLD_DESC = "(Lnet/minecraft/world/IBlockAccess;III)Lnet/minecraft/util/IIcon;";
     private static final String MISSINGNO_DESC = "()Lnet/minecraft/util/IIcon;";
 
+    // These are the targeted calls - a variety of getIcons declared by the Block class
     private static final String GI_WORLD = "getIcon(Lnet/minecraft/world/IBlockAccess;IIII)Lnet/minecraft/util/IIcon;";
     private static final String GI_SIDE_META = "getIcon(II)Lnet/minecraft/util/IIcon;";
     private static final String OBF_SIDE_META = "func_149735_b(II)Lnet/minecraft/util/IIcon;";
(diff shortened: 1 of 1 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/client/model/loading/ModelRegistry.java
```diff
@@ -166,24 +184,43 @@ private void detectAndLoadTextures(GlobalResourceManager manager) {
...
+                        MODEL_LOGGER.error("Block registry contained a non-block or the info mixin failed!");
+                        MODEL_LOGGER.error("Either you won't notice anything, or all JSON models will stop loading...");
+                        infoMixinFailed = true;
+                    }
+                    return;
+                }
+                final var modeled = modeledBlocks.contains(name);
+                modelInfo.nhlib$setModeled(modeledBlocks.contains(name));
+                if (modeled)
+                    MinecraftForgeClient.registerItemRenderer(Item.getItemFromBlock(block), ModelISBRH.INSTANCE);
```
Ruling-0: For safety, worth null checking `getItemFromBlock`?

src/main/java/com/gtnewhorizon/gtnhlib/core/fml/transformers/BlockIconTransformer.java
```diff
@@ -0,0 +1,147 @@
...
+    /// This is what our hook is doing. It assumes that the
+    /// ```java
+    /// // For the IBlockAccess variant
+    /// if (this.nhlib$isModeled) return ModelISBRH.INSTANCE.getParticleIcon(world, x, y, z);
+    /// // For the side:meta variants
+    /// if (this.nhlib$isModeled) return ModelISBRH.INSTANCE.getParticleIcon((Block) this, meta);
+    /// ```
+    @Override
+    public byte[] transform(String name, String transformedName, byte[] basicClass) {
+        final var cr = new ClassReader(basicClass);
```
Ruling-0: Missing null check
```suggestion
    public byte[] transform(String name, String transformedName, byte[] basicClass) {
        if (basicClass == null) return null;
        final var cr = new ClassReader(basicClass);
```

PR #320, opened 2026-03-22
Make ModelISBRH.INSTANCE ThreadLocal to fix race condition

Ruling-0:
ModelISBRH has mutable fields `worldContext` and `itemContext`. `worldContext` especially has race conditions as it may be reset by Angelica rendering threads. This just makes INSTANCE a ThreadLocal, which should fix the issues cleanly.


ah-OOG-ah:
ModelISBRH is already marked as a per-thread ISBRH. What is using it across threads that isn't respecting that?

Ruling-0:
> ModelISBRH is already marked as a per-thread ISBRH. What is using it across threads that isn't respecting that?

The mixin, see discord discussion: https://discord.com/channels/181078474394566657/603348502637969419/1485411672456822794

ah-OOG-ah (approved):

PR #322, opened 2026-04-05
Minor configuration system cleanup

Nikolay-Sitnikov:
Make the system more convenient to use by making `ConfigException` extend `RuntimeException`, fix a bug where string arrays would be incorrectly synced, and remove some outdated Lombok.


Lyfts (approved):
lgtm

PR #328, opened 2026-04-13
Add & color code awareness to font methods

alppp:
add & color code awareness to font width methods

adds support for &#RRGGBB, &g gradients, and single &X codes in sizeStringToWidth
and trimStringToWidth so they're treated as zero-width (same as § codes).

all & handling is gated behind textPreprocessor != null, so without angelica
installed & is just a normal character. also tracks bold state (&l/&r) in the
skip paths so wrap points are correct for bold text.

requires: https://github.com/GTNewHorizons/Angelica/pull/1669

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
@@ -2,6 +2,7 @@ package com.gtnewhorizon.gtnhlib.util.font;
 
 import java.util.function.Function;
 
+import lombok.Setter;
 import net.minecraft.client.gui.FontRenderer;
 
 // Common font rendering utilities that may be better-behaved than vanilla counterparts
@@ -9,14 +10,11 @@ public class FontRendering {
 
     private static final char FORMATTING_CHAR = 167; // §
 
+    @Setter
     private static Function<String, String> textPreprocessor = null;
 
-    public static void setTextPreprocessor(Function<String, String> preprocessor) {
-        textPreprocessor = preprocessor;
-    }
-
     /**
-     * Apply the registered text preprocessor (e.g. &amp;RRGGBB → §x conversion). Returns the input unchanged if no
+     * Apply the registered text preprocessor (e.g. &RRGGBB → §x conversion). Returns the input unchanged if no
      * preprocessor is set.
      */
     public static String preprocessText(String str) {
@@ -175,8 +173,8 @@ public class FontRendering {
                     }
                     break;
                 case '&':
-                    // Skip & color codes as zero-width only when a preprocessor is registered
-                    // (i.e. Angelica is present). Without it, & has no special meaning.
+                    // Skip & color codes as zero-width only when a text preprocessor is registered.
+                    // Without one, & has no special meaning.
                     if (textPreprocessor != null && i + 1 < originalStringLength) {
                         char next = str.charAt(i + 1);
                         if (next == '#' && isHex6(str, i + 2)) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
```diff
@@ -1,12 +1,46 @@
...
     private static final char FORMATTING_CHAR = 167; // §
 
+    private static Function<String, String> textPreprocessor = null;
+
+    public static void setTextPreprocessor(Function<String, String> preprocessor) {
+        textPreprocessor = preprocessor;
+    }
+
+    /**
+     * Apply the registered text preprocessor (e.g. &amp;RRGGBB → §x conversion). Returns the input unchanged if no
```
DeathFuel:
```suggestion
     * Apply the registered text preprocessor (e.g. &RRGGBB → §x conversion). Returns the input unchanged if no
```

src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
```diff
@@ -154,6 +250,38 @@ public static String trimStringToWidth(String str, int trimWidth, boolean revers
                 curBold = determineIfBold(curBold, ch);
             } else if (charWidth < 0) {
                 parsingFormatCode = true;
+            } else if (!reverse && ch == '&' && textPreprocessor != null && i + 1 < str.length()) {
+                // Skip & color codes as zero-width (forward direction only, only when Angelica is present)
```
DeathFuel: How do you know that Angelica is present? `textPreprocessor` can be set by anything, right?

src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
```diff
@@ -1,12 +1,46 @@
...
 import net.minecraft.client.gui.FontRenderer;
 
 // Common font rendering utilities that may be better-behaved than vanilla counterparts
 public class FontRendering {
 
     private static final char FORMATTING_CHAR = 167; // §
 
+    private static Function<String, String> textPreprocessor = null;
+
+    public static void setTextPreprocessor(Function<String, String> preprocessor) {
```
DeathFuel: Consider using Lombok's Setter annotation instead

src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
alppp: fixed

src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
alppp: angelica is the only thing that calls `setTextPreprocessor()`, so if it's non-null angelica set it up. technically anything could call it but nothing else does

src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
alppp: done, using `@Setter` now

src/main/java/com/gtnewhorizon/gtnhlib/util/font/FontRendering.java
DeathFuel: I'm not a fan of this solution tbh, it seems fickle... if there's no better way then please at least document the assumption you're making (that "anything could call it but nothing else does")

DeathFuel (approved):

PR #329, opened 2026-04-15
add colors to /blockstate property values

flamingowrangler2869:
integer: yellow
float: light purple
string: aqua
boolean: green/red


tiffit:
image
enum values should also have a color I think (maybe same as strings)

Roadhog360:
Suggestion: Make it so a property that is not world mutable is red? (as in the property name itself, not the value) And possibly a color for when !SupportsWorld && !WorldMutable, which would be essentially a pure function.

Just a suggestion though, not a requested change because it is not essential to me merging this PR.

Roadhog360:
A review is coming soon btw

flamingowrangler2869:
Alr, i'll get on it, been slightly busy lately xd

flamingowrangler2869:
java_lIPUMClUa5

Roadhog360:
I forgor

RecursivePineapple (approved):

PR #330, opened 2026-04-17
add title/subtitle api, overlay, and /title command

alppp:
add a client-side title api for any mod to display title/subtitle overlay text. if a renderer (e.g. hodgepodge) registers itself, it handles full rendering with scale, fade, and subtitle. 
otherwise setTitle falls back to the vanilla action bar overlay.

- setTitle / setSubtitle (IChatComponent)
- setTimes / clear / reset (1.21 semantics)
- registerRenderer for full overlay rendering
- tick() decrements timer per client tick
- @Getter on volatile fields (lombok)
- @SideOnly(Side.CLIENT)


Alexdoru:
how does that work when a dedicated server need to send text to a client?

I forget is the title/subtitle implemented in 1.7.10 or did it only appear in 1.8?

alppp:
migrated from hodgepodge to there : 
examples : 
image

image
also 1.7.10 has only "actionbar" about receiving, it only receives those that exists, you cannot do 

`/title title show something`

 you always need to action a player or you can do `@a` (which will send only active players) 

`title @a title hello world`

Alexdoru:
can you update the title of this PR ?

Lyfts (approved):
Tested and works well, gj.

PR #334, opened 2026-04-26
Add @Config.{RequiresMod, RequiresModOr and RequiresModAnd} to the config system

slprime:
Example: 
```java
    @Config.Comment("Only visible in GUI if 'Hodgepodge' mod is loaded.")
    @Config.RequiresMod("hodgepodge")
    public static boolean hodgepodgeIntegration = true;

    @Config.Comment("Only visible in GUI if 'nei' or 'jei' is loaded.")
    @Config.RequiresMod({ "nei", "jei" })
    public static boolean itemSearchIntegration = false;

    @Config.Comment("This entire category is hidden unless 'thaumcraft' is loaded.")
    @Config.RequiresMod(value = { "thaumcraft", "thaumicenergistics" }, mode = Config.RequiresMod.Mode.AND)
    @Config.Order(1)
    public static ThaumcraftCategory thaumcraft = new ThaumcraftCategory();

    public static class ThaumcraftCategory {

        @Config.Comment("Enable Thaumcraft research tweaks.")
        public boolean researchTweaks = true;

        @Config.Comment("Only visible if 'thaumicbases' is also present.")
        @Config.RequiresMod("thaumicbases")
        public boolean thaumicBasesCompat = false;
    }
```

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/ExampleConfig.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/ExampleConfig.java
@@ -73,11 +73,11 @@ public class ExampleConfig {
     public static boolean hodgepodgeIntegration = true;
 
     @Config.Comment("Only visible in GUI if 'nei' or 'jei' is loaded.")
-    @Config.RequiresMod({ "nei", "jei" })
+    @Config.RequiresModOr({ "nei", "jei" })
     public static boolean itemSearchIntegration = false;
 
     @Config.Comment("This entire category is hidden unless 'thaumcraft' is loaded.")
-    @Config.RequiresMod("thaumcraft")
+    @Config.RequiresModAnd({ "thaumcraft", "thaumicenergistics" })
     @Config.Order(1)
     public static ThaumcraftCategory thaumcraft = new ThaumcraftCategory();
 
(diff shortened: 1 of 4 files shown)


Alexdoru:
should probably make it explicit that it's only for the GUI and that if you put multiple mods it's a OR operation, why not also add a AND

`@Config.GUI.RequiresModOr({ "nei", "jei" })`
`@Config.GUI.RequiresModAnd({ "nei", "jei" })`

Lyfts:
Instead of the now 3 separate annotations why not just use an Enum for the mode and add that to the `@Config.RequiresMod` annotation?

slprime:
> Instead of the now 3 separate annotations why not just use an Enum for the mode and add that to the `@Config.RequiresMod` annotation?

Good point, but I'd prefer to keep the annotations separate. A few reasons:

**Clarity at the point of use.** `@RequiresModAnd({"modA", "modB"})` is self-documenting — the logic is encoded in the annotation name itself. With an enum discriminator, the reader has to parse two attributes to understand intent:

```java
@RequiresMod(value = {"modA", "modB"}, mode = Mode.AND) // less obvious
```

**The common case gets worse.** The single-mod usage `@RequiresMod("mod")` is the most frequent. With a combined annotation you'd either force an array-of-one syntax (`value = {"mod"}`) or add a separate `String value()` alongside `String[] values()` — which makes the API more confusing, not less.

Lyfts:
Sure thing, but as to your second point you can use `@RequiresMod("mod")` even if the required value is a `String[]`. We do that all the time for `@Config.Comment`

slprime:
> Sure thing, but as to your second point you can use `@RequiresMod("mod")` even if the required value is a `String[]`. We do that all the time for `@Config.Comment`

done

Lyfts (approved):

PR #338, opened 2026-05-01
Add world deletion event mixin

Algent:
This PR create a `WorldDeletionEvent` event posted to the Minecraft Forge event bus.

The `worldName` string is the name of the currently selected world, which is also the folder name for that save.

This is an alternative approach to https://github.com/GTNewHorizons/Hodgepodge/pull/864 
It will allow to handle the cleanup in each separate mods.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
@@ -4,9 +4,21 @@ import cpw.mods.fml.common.eventhandler.Event;
 
 public final class WorldDeletionEvent extends Event {
 
-    // Not a full path, name of just the directory as it is in .minecraft\saves\
     public final String worldName;
 
+    /**
+     * {@link WorldDeletionEvent} is posted on the {@link net.minecraftforge.common.MinecraftForge#EVENT_BUS} a world
+     * deletion is initiated through the world selection menu in the client and the world directory name has been
+     * resolved.
+     * </p>
+     * <p>
+     * {@link #worldName} String contains the name of the world directory as it exists under {@code .minecraft/saves},
+     * this is not a full or absolute filesystem path.
+     * </p>
+     * <p>
+     * This event is not {@link cpw.mods.fml.common.eventhandler.Cancelable}.
+     * </p>
+     */
     public WorldDeletionEvent(String worldName) {
         this.worldName = worldName;
     }
--- a/src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
@@ -1,5 +1,8 @@
 package com.gtnewhorizon.gtnhlib.client.event;
 
+import net.minecraftforge.common.MinecraftForge;
+
+import cpw.mods.fml.common.eventhandler.Cancelable;
 import cpw.mods.fml.common.eventhandler.Event;
 
 public final class WorldDeletionEvent extends Event {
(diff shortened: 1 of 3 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinGuiSelectWorld.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Mixin(value = GuiSelectWorld.class)
+public class MixinGuiSelectWorld extends GuiScreen {
+
+    @Shadow
+    private List field_146639_s;
+
+    @Inject(method = "confirmClicked", at = @At("HEAD"))
+    public void onConfirmClicked(boolean result, int id, CallbackInfo ci) {
+        String fileName = ((SaveFormatComparator) this.field_146639_s.get(id)).getFileName();
```
mitchej123: does `getFileName` return just the filename, or the full file path?

Algent: It return just the filename.
To get full path you would have to do `Paths.get(Minecraft.getMinecraft().mcDataDir, "saves", worldName)`

src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
```diff
@@ -0,0 +1,14 @@
+package com.gtnewhorizon.gtnhlib.client.event;
+
+import cpw.mods.fml.common.eventhandler.Event;
+import lombok.Getter;
+
+public final class WorldDeletionEvent extends Event {
+
+    @Getter
+    private final String worldName;
```
Alexdoru: usually event fields are simply public there is no getter

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinGuiSelectWorld.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.gtnewhorizon.gtnhlib.client.event.WorldDeletionEvent;
+
+@Mixin(value = GuiSelectWorld.class)
+public class MixinGuiSelectWorld extends GuiScreen {
+
+    @Shadow
+    private List field_146639_s;
+
+    @Inject(method = "confirmClicked", at = @At("HEAD"))
```
Alexdoru: the injection is wrong you should not inject at HEAD, you need to inject in the deleteWorldDirectory to capture the argument and only run when this code runs

Alexdoru (changes requested):

src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
```diff
@@ -0,0 +1,14 @@
+package com.gtnewhorizon.gtnhlib.client.event;
+
+import cpw.mods.fml.common.eventhandler.Event;
+import lombok.Getter;
+
+public final class WorldDeletionEvent extends Event {
+
+    @Getter
```
Alexdoru: whatever this string is you should add documentation of what it is, is it the file filepath is it just the directory name , if its a full path is it `/` seoarated?

src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
Algent: Understood

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinGuiSelectWorld.java
Algent: Changed into @ModifyArg as suggested

src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
Algent: I added a comment above, hopefully it's clear enough to avoid confusion of anyone wanting to use it later.

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinGuiSelectWorld.java
```diff
@@ -0,0 +1,26 @@
...
+public class MixinGuiSelectWorld extends GuiScreen {
+
+    @ModifyArg(
+            method = "confirmClicked",
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lnet/minecraft/world/storage/ISaveFormat;deleteWorldDirectory(Ljava/lang/String;)Z"),
+            index = 0)
+    private String onDeleteWorldDirectoryCalled(String worldName) {
+        MinecraftForge.EVENT_BUS.post(new WorldDeletionEvent(worldName));
```
Alexdoru: i'd put that in a try/catch

src/main/java/com/gtnewhorizon/gtnhlib/client/event/WorldDeletionEvent.java
```diff
@@ -0,0 +1,13 @@
+package com.gtnewhorizon.gtnhlib.client.event;
+
+import cpw.mods.fml.common.eventhandler.Event;
+
+public final class WorldDeletionEvent extends Event {
+
+    // Not a full path, name of just the directory as it is in .minecraft\saves\
```
Alexdoru: that's not how we do javadoc

Algent: Is it better ?

Alexdoru (approved):

PR #341, opened 2026-05-03
Fix ModelISBRH.INSTANCE ThreadLocal incompatibility

ABKQPO:
beacause #320 
tested with OKBackpack on exp104


Caedis:
Can you describe the issue and fix with more details when dealing since this is related to ASM?

Alexdoru (approved):

Alexdoru:
> Can you describe the issue and fix with more details when dealing since this is related to ASM?

get good

PR #343, opened 2026-05-06
Add Model overrideIcon

Shigure-Ruiseki:
This PR picks up from the [fix-cracks](https://github.com/GTNewHorizons/GTNHLib/tree/fix-cracks) branch

https://github.com/user-attachments/assets/ac60ea12-5094-4f5c-9bd7-067aaaf1fece

https://github.com/user-attachments/assets/84826d37-5ac1-486f-8133-1c16a7144491


ah-OOG-ah:
Nice, someone picked up the branch. 

Does this properly rotate, scale, and tile the cracks on oddly-shaped models? I used the lectern for reference previously, as it combines slopes, partial faces, and faces extending outside the 1-block bounding box.

Shigure-Ruiseki:
@ah-OOG-ah fixed?

https://github.com/user-attachments/assets/c27cc524-1b00-4829-84d6-c375c0c8f5c9

ah-OOG-ah:
Look at a lectern breaking in modern (I think any version with a lectern will do) - note how this PR currently just places the break texture aligned with the normal face's texture.

Modern always has the block breaking texture aligned to the block grid as long as the face is - see the picture below. For faces parallel to the grid but inset the texture is simply inset and cropped to the quad, while it only gets applied in the same manner as the quad's normal texture if the quad is at an angle, as the top face of the lectern is.
image

Shigure-Ruiseki:
> Modern always has the block breaking texture aligned to the block grid as long as the face is - see the picture below. For faces parallel to the grid but inset the texture is simply inset and cropped to the quad, while



> Look at a lectern breaking in modern (I think any version with a lectern will do) - note how this PR currently just places the break texture aligned with the normal face's texture.
> 
> Modern always has the block breaking texture aligned to the block grid as long as the face is - see the picture below. For faces parallel to the grid but inset the texture is simply inset and cropped to the quad, while it only gets applied in the same manner as the quad's normal texture if the quad is at an angle, as the top face of the lectern is. image

Currently, I still haven't found the optimal way to do this, so if I can, next time I will find a way to do it like modern.
This time pr just added override block to make it work well enough.
I prioritize making Item Models

ah-OOG-ah (approved):

ah-OOG-ah (dismissed):
@UltraProdigy needs to compress the images, since this is coming from a fork

UltraProdigy (approved):

PR #344, opened 2026-05-06
redo ITeamData#copyData signature

tiffit:
The previous copyData didn't really make any sense. The new changes make it similar to mergeData:
- Called on the destination team data
- Passes in prevTeamData
- Doesn't return anything

Currently, no mods use this library and I would like to get this change in before any do


cubefury (approved):
nice

PR #346, opened 2026-05-07
Fix team names with spaces not working due to greedystring

### LINKED ISSUE #345: Team names with spaces in them have issues

cubefury

### Mod Version

0.9.61

### Bug Report

GTNHTeams that have spaces in team names don't work correctly. I'll push a fix soon tm unless someone else gets to it first.

### Java Version

Java 17

devenv

### Final Checklist

cubefury:
This PR aims to Fix #345.

Team names with spaces in them had some issues. For simplicity, if team names have spaces they must now have quotes around them in order to get caught by the quotable string arguments. 
Single-word teamnames don't require quotes still.

Some of the commands like the force-disband or self-rename technically don't need the quotes, but I feel like this might just become a source of user confusion if we allow that.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamAdminCommand.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamAdminCommand.java
@@ -38,8 +38,8 @@ public class TeamAdminCommand {
                         })
                         .then(
                                 literal("rename").then(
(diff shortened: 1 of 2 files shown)


tiffit (changes requested):
If we are going to require quotes for any team name with spaces, we should update the message sent on invite to include quotes. Though, it might be easier to just use greedy string for `/gtnhteam accept` and `/gtnhteam deny`

tiffit (approved):

PR #350, opened 2026-05-11
Data Component

Shigure-Ruiseki:
This use for item model can read item properties like block model with blockstate
image


RecursivePineapple:
This new system seems kinda redundant, block properties already support item stacks. It looks like this is nearly a verbatim copy of it.

Shigure-Ruiseki:
> This new system seems kinda redundant, block properties already support item stacks. It looks like this is nearly a verbatim copy of it.

I need this system to implement the item JSON model (not the item block model), so I cannot use block state properties.

PR #351, opened 2026-05-14
Fix teams logic and introduce playerutils

cubefury:
1. Fixed some logic in teams code where a team which was merged into another/disbanded would have some references still lingering.
2. Implemented and swapped to ServerPlayerUtils classes - we're going to move all resolution of player uuid and name strings to this place, to avoid confusion like "should we use getDisplayName() or getCommandSenderName() or getGameProfile().getName()"

Tested in in dev env. 

ps Please forgive the commit history - it's a mess because somehow some changes weren't stashed while working on the teams gui and a bunch of them made it into a commit.


cubefury:
Some of the methods in the serverplayerutils and clientplayerutils aren't used atm, but I'm prepositioning for their use in teams gui and other places as people see fit.

tiffit (approved):
[code looks good](https://en.wikipedia.org/wiki/Gravel_cycling)

PR #352, opened 2026-05-15
Include more data client-side for UI display

cubefury:
PR description last updated: 22 May

<resolved: we're syncing the teamdata when the team members change in any way> Before reviewing the changes, I am seeking advice on whether we need to resync the full teamdata (eg. vending machine wallet) when: a) player accepts team invite, b) team merge is accepted/force team merge and c) team is disbanded by admin and players are dumped into their solo teams, or if this will be handled some other way. Currently I have put them in, but please let me know if it's not necessary and I'll take the teamdatasyncs out.

2. <update: this has been taken out altogether as it is too expensive to sync all player name data everytime a player logs in on larger servers> PlayerDataSync: Syncing UUIDs over to the clientside is scrapped, and instead we will just stick to the player name strings.

3. Fixed a couple places where we forgot to sync data over.

I tested the changes in devenv with debug output to dump the full teamlist and their members to ensure they're working. All the commands and admin commands look like they're still working. Sorry for the wall of text!


cubefury:
Oh also team names now have a max length cuz mui2's textfield wants one, but surely no one's team will be 10k characters long...

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamManager.java
```diff
@@ -41,6 +42,10 @@ public static boolean isTeamNameValid(String name) {
         return true;
     }
 
+    public static Map<UUID, Team> getTeamMap() {
+        return TEAM_MAP;
+    }
```
tiffit, line 47: We shouldn't expose the TEAM_MAP through a public API. Modifications to this map could cause corrupted data and exposing it means anyone can do anything they want with it.

tiffit (changes requested):

cubefury:
After some discussion, I will be reworking this to implement a request+response system to only send data which is necessary

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamManager.java
cubefury: Good point, if this method has to stay after the rest of the changes I'll make it return an unmodifiable copy or smth instead

cubefury:
Changes since:

1. I removed playerdatasync for now, the online list would be too costly to sync to the client everytime someone logs in on large servers. If we want to sync just the changes to the UsernameCache we'll need to mixin(?) to the onPlayerLogin code and check if there are any changes to the server-side cache before syncing.
2. TeamInfoSync has been reverted and now just tracks an additional playerrole for teams GUI.
3. TeamActions unifies some duplicate code from TeamCommand and TeamAdminCommand, and will also be used in the teams GUI later on.

Tested in SP and MP in devenv

src/main/java/com/gtnewhorizon/gtnhlib/eventhandlers/ServerEventHandler.java
```diff
@@ -20,14 +16,9 @@ public class ServerEventHandler {
 
     @SubscribeEvent(priority = EventPriority.LOW)
```
tiffit: Any reason why the priority is LOW? Ideally its normal so other mods can easily run stuff before/after a team is created on login, such as migrating data

src/main/java/com/gtnewhorizon/gtnhlib/util/ServerPlayerUtils.java
```diff
@@ -42,8 +40,11 @@ public static Map<UUID, EntityPlayerMP> getOnlinePlayers() {
     }
 
     public static UUID getPlayerUUID(String name) {
```
tiffit, line 45: could add a @Nullable to the return type

src/main/java/com/gtnewhorizon/gtnhlib/util/ClientPlayerUtils.java
```diff
@@ -1,37 +1,26 @@
...
         return Minecraft.getMinecraft().thePlayer.sendQueue.playerInfoList.stream().map(info -> info.name)
                 .collect(Collectors.toSet());
     }
 
-    public static UUID getPlayerUUID(String name) {
-        synchronized (clientPlayerMap) {
-            return clientPlayerMap.get(name);
-        }
+    public static String getCurrentPlayerName() {
+        return Minecraft.getMinecraft().thePlayer.getCommandSenderName();
```
tiffit: `thePlayer` can be null. If it is, might be good to fallback to `Minecraft.getMinecraft().getSession().getUsername()` so this method can be used from the main menu if someone else wants to use it

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamCommand.java
```diff
@@ -527,11 +371,26 @@ private static int executeMergeDeny(ICommandSender sender, String sourceTeamName
...
 
-        ChatComponentText sourceComponent = new ChatComponentText(source.getTeamName());
-        sourceComponent.getChatStyle().setColor(EnumChatFormatting.GOLD);
-        return success(sender, "gtnhlib.chat.teams.message.merge_denied", sourceComponent);
+        return Command.SINGLE_SUCCESS;
+    }
+
+    private static int executeDisband(ICommandSender sender) {
+        EntityPlayer player = TeamCommandsUtils.asPlayer(sender);
+        UUID playerId = player.getUniqueID();
```
tiffit: this will cause an NPE if player is null. Needs to be moved after the check below

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamActions.java
```diff
@@ -0,0 +1,317 @@
...
+import net.minecraft.util.ChatComponentTranslation;
+import net.minecraft.util.EnumChatFormatting;
+
+import com.gtnewhorizon.gtnhlib.network.NetworkHandler;
+import com.gtnewhorizon.gtnhlib.network.teams.TeamDataSync;
+import com.gtnewhorizon.gtnhlib.util.ServerPlayerUtils;
+
+public class TeamActions {
+
+    public static void onRename(Team team, String oldName, String newName, boolean adminAction, ICommandSender admin) {
```
tiffit: `ICommandSender admin` could get the @Nullable annotation. Same for all the other ones that take admin as an arg

src/main/java/com/gtnewhorizon/gtnhlib/teams/TeamActions.java
```diff
@@ -0,0 +1,317 @@
...
+                colorChatComponent(EnumChatFormatting.GOLD, teamName));
+        notice.getChatStyle().setColor(EnumChatFormatting.RED);
+
+        for (UUID uuid : members) {
+            String name = ServerPlayerUtils.getPlayerName(uuid);
+            Team newTeam = TeamManager.getOrCreateTeam(name, uuid);
+            TeamManager.transferTeamData(team, newTeam, uuid, TeamDataTransferReason.JoinedNewTeam);
+            TeamManager.forEachOnlineTeamMember(newTeam, member -> {
+                NetworkHandler.instance.sendTo(TeamNetwork.createTeamInfoSyncPacket(member.getUniqueID()), member);
+                TeamNetwork.sendPlayerAllTeamData((EntityPlayerMP) member, newTeam);
```
tiffit: The `(EntityPlayerMP)` cast is not needed

tiffit (changes requested):
some minor things, looks good otherwise

src/main/java/com/gtnewhorizon/gtnhlib/eventhandlers/ServerEventHandler.java
cubefury: I moved it to low because we had the playerdatasync for the cached usernames, and this had to take place after. But now that playerdatasync is no longer a thing I'll move it back to the normal priority

tiffit (approved):

PR #353, opened 2026-05-16
Angelica patch

sisyphussy:
- Make CallbackTessellator's Callback global state (Every CallbackTessellator uses the same Callback, allowing for full control of the pipeline) & allow vertex mutations
- Add VertexTransformCallback for easy transforms (will be used in display lists, can be used for baked models too)
- Make GlyphReplacements a Char2Char map
- Deprecate some more methods
- Make VertexFormat less extensible

New angelica will require this version, but older angelica will work with this version just fine.


src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
```diff
@@ -64,20 +65,57 @@ public int draw() {
...
+        }
+        if (other.hasBrightness) {
+            this.setBrightness(other.brightness);
+        }
+        if (other.hasNormals) {
+            final int normal = other.normal;
+            byte b0 = (byte) (normal & 0xFF);
+            byte b1 = (byte) ((normal >> 8) & 0xFF);
+            byte b2 = (byte) ((normal >> 16) & 0xFF);
+            this.setNormal(b0, b1, b2);
```
mitchej123: Is this right? There are no byte variants of setNormal() that I saw

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
```diff
@@ -64,20 +65,57 @@ public int draw() {
...
-        this.drawMode = tessellator.drawMode;
+    public final void syncVanillaTessellator() {
+        syncTessellator(TessellatorManager.getVanillaTessellator());
+    }
+
+    public final void syncTessellator(Tessellator other) {
+        if (!other.isDrawing || other.vertexCount == 0) return; // Not drawing, nothing to sync
+
+        // Cannot merge when both tessellators are drawing with different draw mode
+        if (this.isDrawing && other.drawMode != this.drawMode) {
```
mitchej123, line 76: if not drawing, it doesn't sync the drawMode.  Intentional?

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
mitchej123: Why not directly set this.normal?

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
```diff
@@ -64,20 +65,57 @@ public int draw() {
...
+        }
+        if (other.hasNormals) {
+            final int normal = other.normal;
+            byte b0 = (byte) (normal & 0xFF);
+            byte b1 = (byte) ((normal >> 8) & 0xFF);
+            byte b2 = (byte) ((normal >> 16) & 0xFF);
+            this.setNormal(b0, b1, b2);
+        }
+
         this.format = preDefinedFormat != null ? preDefinedFormat : getOptimalVertexFormat();
```
mitchej123, line 121: What will happen if there's already stuff in this and the format changes?

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
```diff
@@ -99,7 +137,6 @@ public void reset() {
         this.isColorDisabled = false;
 
         this.format = null;
-        this.preDefinedFormat = null;
```
mitchej123, line 102: Not clear why this is not being cleared now?  Seems a lot of things guard on it and return if not null?

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/VertexCallbackManager.java
```diff
@@ -0,0 +1,41 @@
...
+        tessellatorCallbacks[0] = DEFAULT;
+        callback = DEFAULT;
+    }
+
+    private static void updateCallback() {
+        callback = tessellatorCallbacks[callbackIndex];
+    }
+
+    public static void pushCallback(TessellatorCallback callback) {
+        if (++callbackIndex >= DIRECT_TESSELLATOR_STACK_DEPTH) {
```
mitchej123, line 27: unless we crash, this will hit an IOOBE error in popCallback().  Should check and then increment, or decrement before throwing. 

If not recoverable at all, suppose its fine.

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/CallbackTessellator.java
```diff
@@ -1,23 +1,13 @@
...
-        super(Tessellator.byteBuffer);
-        this.drawCallback = callback;
-    }
-
-    public CallbackTessellator(ByteBuffer initial, DirectDrawCallback callback) {
-        super(initial);
-        this.drawCallback = callback;
-    }
-
     public CallbackTessellator(ByteBuffer initial) {
```
mitchej123, line 12: Where can this be instantiated from? The general methods aren't adding a callback, but it's unconditionally popped from `onRemovedFromStack`

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/MatrixHelper.java
```diff
@@ -0,0 +1,49 @@
...
+import org.joml.Matrix3f;
+import org.joml.Matrix4f;
+import org.joml.Vector3f;
+import org.joml.Vector4f;
+
+import com.gtnewhorizon.gtnhlib.client.model.NormalHelper;
+
+public final class MatrixHelper {
+
+    private static final Matrix4f IDENTITY = new Matrix4f();
```
mitchej123: might be worth making this a Matrix4fc so it can't be mutated

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/tessellator/VertexCallbackManager.java
```diff
@@ -0,0 +1,41 @@
...
+
+import static com.gtnewhorizon.gtnhlib.client.renderer.TessellatorManager.DIRECT_TESSELLATOR_STACK_DEPTH;
+
+import com.gtnewhorizon.gtnhlib.client.renderer.TessellatorCallback;
+
+public final class VertexCallbackManager {
+
+    public static final TessellatorCallback DEFAULT = new TessellatorCallback() {};
+
+    public static TessellatorCallback callback;
```
mitchej123: Any reason this needs to be public?

src/main/java/com/gtnewhorizon/gtnhlib/util/font/GlyphReplacements.java
```diff
@@ -2,21 +2,37 @@
...
+public final class GlyphReplacements {
 
     /**
      * A mapping from custom glyphs to the values they are meant to replace (which may be null for novel glyphs or when
      * the original comes from outside the BMP).
      */
+    @Deprecated
     public static final HashMap<String, String> customGlyphs = new HashMap<>(64);
 
+    private static final Char2CharOpenHashMap customGlyphsNew = new Char2CharOpenHashMap();
```
mitchej123, line 16: Does this need a default sentinel value? -1?  or will things handle 0/NUL properly?

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/CallbackTessellator.java
```diff
@@ -32,14 +22,122 @@ public CallbackTessellator(ByteBuffer initial, boolean deleteAfter) {
...
+            final int result = super.draw();
+            reset();
+            return result;
+        }
+
+        return 0; // onDraw returns false -> no draw happening
+    }
+
+    @Override
+    public void startDrawing(int drawMode) {
```
mitchej123, line 36: Most Tesellators throw if already drawing, this used to (because it delegated to the parent); now it doesn't.  Intentional?

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
sisyphussy: I could've sworn I added a byte variant... Main reason I want it to go through setNormal is so that CallbackTessellator can mutate it.

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
sisyphussy: setColor, setBrightness etc should automatically attempt to fix the buffer format via fixBufferFormat

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
sisyphussy: I noticed a bug where the behavior of DirectTessellator and Tessellator deviated because Tessellator calls reset() in startDrawing(), when DirectTessellator didn't (caused some issues with draw merging). I added a reset() to startDrawing to fix that, but that would mean following code would break:

```
tessellator.setVertexFormat(format);
tessellator.startDrawing();
```

Since preDefinedFormat is something that should only be used in the context of using the direct tessellator to populate a VBO (which is also why I now regret not splitting it up into another Tessellator), onRemovedFromStack is perfectly fine here.

src/main/java/com/gtnewhorizon/gtnhlib/util/font/GlyphReplacements.java
sisyphussy: chars are unsigned, default return value will be 0. Will add a comment to clarify.

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/tessellator/VertexCallbackManager.java
sisyphussy: I created a new tessellator package to not clutter the renderer package further, so they are in different packages

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/DirectTessellator.java
```diff
@@ -64,20 +65,62 @@ public int draw() {
...
+                    new IllegalStateException());
+            return;
+        }
+
+        // If this tessellator has emitted a vertex, try to merge their data together
+        // Really scuffed, but should cover some rare edge-cases
+
+        this.isDrawing = true;
+        this.drawMode = other.drawMode;
+        this.vertexCount = this.vertexCount + other.vertexCount;
```
mitchej123: This bumps the vertexCount, before adding the new verts, anything that calls `fixBufferFormat` is going to either crash or mess with garbage data via `for (int i = 0; i < vertexCount; i++)`

src/main/java/com/gtnewhorizon/gtnhlib/client/renderer/VertexCallbackManager.java
mitchej123: setDirectTessellator also has this problem

mitchej123 (approved):

PR #354, opened 2026-05-16
Add fast WavefrontObject loader

sisyphussy:
- Adds WavefrontVBOBuilder
  - Unnecessarily optimized parser that doesn't take up any ram and directly uploads to vram. Also has the benefit of having a faster 1st render time (matters a lot for items while browsing NEI)
  - Does NOT replace forge's WavefrontObject because it doesn't support groups.
 - Adds CCWavefrontBuilder 
   - Still WIP, currently lacking support for some stuff, but also has the same benefits (no vram + faster render times)


mitchej123 (approved):

Alexdoru:
how does this work with ?
https://github.com/GTNewHorizons/Hodgepodge/pull/850
https://github.com/GTNewHorizons/Hodgepodge/pull/851

sisyphussy:
I've profiled it, it's faster than the optimization PR

PR #356, opened 2026-05-17
add u (drop shadow) to isValidAmpCode

alppp:
adds &u to isValidAmpCode so width/size/trim methods skip &u as zero-width.

part of the drop shadow feature. &u toggles per-segment shadow, &u&#RRGGBB sets custom shadow color.


sisyphussy (approved):
+2 -1

PR #363, opened 2026-05-22
LivingEquipmentChangeEvent init detection

### LINKED ISSUE #361: `LivingEquipmentChangeEvent` has no way to detect if it's being fired through entity initialization

Roadhog360

### Mod Version

0.10.2

### Bug Report

```java
	@SubscribeEvent
	public static void handleArmorSounds(LivingEquipmentChangeEvent event) {
		if ((event.entity instanceof EntityPlayer player && !SpectatorMode.isSpectator(player)) && event.getSlot() != 0) {
			ItemStack from = event.getFrom();
			ItemStack to = event.getTo();
			String sound = ArmorSoundsRegistry.getEquipSound(to);
			if (((from == null && to != null) || (from != null && to != null && from.getItem() != to.getItem())) && sound != null) {
				player.worldObj.playSoundAtEntity(player, sound, 1, 1);
			}
		}
	}
```

~~On the previous version I was on, `0.8.43-pre` I used the `LivingEquipmentChangeEvent` to simplify armor equip sound logic. Before, the event would not fire on world join. However after updating to the latest version as of writing this, `0.10.2`, it now additionally fires the event on world join.~~ Observing the issue with breakpoints, it appears the event ~~now~~ fires on any equipment slot that has an item in it during entity initialization. So if I am holding something in my hand or wearing any item, the respective slot (0, 1, 2, 3, 4) fires the event.

I am assuming this is unintentional ~~as it would break previous behavior that expected the event to only fire when the item actually changes~~. Additionally, the name LivingEquipment***Change***Event implies the event is supposed to fire on changes, and joining a world is not a change.

~~Weirdly enough I see no changes to GTNH's event code between `0.8.43-pre` and now so I am not really sure why this started happening. I propose that if you want to keep this behavior that there be some way to make the initial pass detectable so it can be easily filtered out from other passes. For now I've just checked if the ticksExisted on the entity is below 20 but this is a very hacky solution.~~

## EDIT: This probably is not a regression in a later update, crossed out anything implying it is.
I thought this after double checking the code, and while making the PR noticed the behavior I described in the now strikethrough text seemed inconsistent, likely due to fix mods fixing "minecart ear blasting" by muting sounds on the loading screen. However I still consider it an issue the event exposes no proper way to detect this, so I've moved forward with a PR to expose the initial runs regardless.  My PR adds a way to properly detect this, solving my problem. With my PR, my issue can be closed as all I have to do is check `event.initial` being false before running the code.

### Java Version

Java 17

N/A

### Final Checklist

Roadhog360:
Added a proper way to detect when an entity is being initialized in LivingEquipmentChangeEvent.

Fixes #361 by adding proper detection while keeping current behavior for existing code.

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/client/event/LivingEquipmentChangeEvent.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/client/event/LivingEquipmentChangeEvent.java
@@ -12,6 +12,7 @@ public class LivingEquipmentChangeEvent extends LivingEvent {
     private final int slot;
     private final ItemStack from;
     private final ItemStack to;
+    private final boolean initial;
 
     /**
      * {@link LivingEquipmentChangeEvent} is fired when the Equipment of an Entity changes. <br>
@@ -29,7 +30,7 @@ public class LivingEquipmentChangeEvent extends LivingEvent {
      * </ul>
      * {@link #from} contains the {@link ItemStack} that was equipped previously. <br>
      * {@link #to} contains the {@link ItemStack} that is equipped now. <br>
-     * <br>
+     * {@link #initial} is {@link Boolean#TRUE} if this entity is firing this event through its first tick. <br>
      * This event is not {@link cpw.mods.fml.common.eventhandler.Cancelable}. <br>
      * <br>
      * This event does not have a result. {@link HasResult} <br>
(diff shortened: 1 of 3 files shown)


FourIsTheNumber:
Just read the issue - it has definitely always fired on world join, and this mirrors the behavior in the modern forge event as well. I think I even remember the modern pr specifying it. This is definitely a useful change though.

FourIsTheNumber (approved):

Roadhog360:
Huh, I wonder why I never had my armor equip sounds play in my dev env until I updated from that same day. I guess my code is just not written to consistently handle that state, or perhaps it was because of rejoins AFTER the first join? Idk, but now that I think about it you're probably right and there was some other factor that made my sound not play before, and the updating was merely a coincidence. Maybe it is related to fix mods that fix "minecart ear blasting" by muting sounds during world load?

Either way, a proper flag rather than a hacky ticksExisted check is definitely something I would like to switch to using.

Roadhog360:
I have updated my issue to strikethrough anything implies this is a regression. However I still left the issue open as I do believe something like this is important for an event of this nature to expose, especially considering what its name is.

PR #364, opened 2026-05-23
Put BlockPropertyRegistry.getGenericInterfaces() into try/catch

JackOfNoneTrades:
`BlockPropertyRegistry.getGenericInterfaces()` crashes on optional interfaces when one of them doesn't resolve. A proper fix would be going over generic interfaces with ASM and do the checks there, but IG this is ok as a first fix.


JackOfNoneTrades:
I discovered this crash when running version (latest as of writing) 1.4.2 of [OK Backpack](https://www.curseforge.com/minecraft/mc-mods/okbackpack/files/8076838) with no Angelica present. The backpack `@Optionally` implements an Angelica interface, and with current NHlib it crashes.

mitchej123 (changes requested):
I don't think this is the correct fix.  Chatting in discord

JackOfNoneTrades:
Fixed by mitch in hodgepodge.

PR #368, opened 2026-05-25
Title config optimizations

sisyphussy:
- Remove volatile (the packets are handled on the main thread)
- Add VertexTransformer for particle system (batches all particles into a single draw call)
- Replace mixins with dynamically registered Forge events


alppp (changes requested):
the event-based approach is cleaner than the mixin and the particle batching is great idea that i didnt think first. few things to fix though:

1) `glDisable(GL_BLEND)` was removed before the final `glPopMatrix` in the render method. the old mixin had it there. this leaks blend state into whatever renders after the title overlay since `glPopMatrix` only restores the matrix stack, not enable states

2) minor: in TitleParticleSystem.render(), if all particles are item-type (PARTICLE_ITEM_CONFETTI), `startDrawingQuads()`/`draw()` runs with zero vertices tessellator handles it fine but the texture state toggles still happen for nothing

alppp (approved):

PR #371, opened 2026-05-26
Support major.minor.patch version format in PackVersion

Eldrinn-Elantey:
PackVersion only accepted major.minor, causing IllegalArgumentException for packs using 3-part versions like 0.0.1. Added optional patch component (defaults to 0), comparison updated accordingly.


Ethryan (approved):
Lgtm

Issue #243, opened 2025-12-26
JSON models unimplemented features


Roadhog360:
This does not use an issue template because this isn't necessarily meant to be presented as an issue or a bug, but rather a simple documentation of what's missing from the JSON model system. I am not in charge of this repository but I do think it would be a good idea to report here when something is missing or not fully implemented so anyone working on it have a clear, unified list on what they can work on.

List of things that JSON models don't support:
## Blockstate files
- `z` rotation field is not implemented (added as of 1.21.11)
- `uvlock` field is not supported
- Multipart models are not supported
- Random model selection is not supported
  - Selection is "random" based on world coords. I have a feeling blocks also have a way to specify how the random model is chosen but I have no proof of that.
- Blockstate system is not currently hooked up to the model definitions, and uses `meta` "states"
  - Note that item frames are treated as a block, and also pass in a "fake state", `map`. This is `false` if the item frame is not holding a map, and `true` if it is. Item frames seem to be rendered with json data instead of a hardcoded entity model.
- Blockstate files cannot override blocks that don't already use the system.
  - Can be fixed fairly trivially by tracking the names of blockstate files in resource packs, then using `@WrapMethod` on `Block#getRenderType` to override its render type if a model is present in any resource pack to override the ID. I think doing it using `@WrapMethod` is the best way since any mod checking the ID (for some reason) also gets this value as well.
## Model files
- Item models are not supported
  - We currently assume what the item model should be via the default blockstate model, but this makes it impossible to implement a different inventory model for blocks. Since the json system seems to focus on blocks anyways, "meeting in the middle" by only supporting block item jsons should be "good enough". Block items refer to what block model they want (and I think on a few edgecases, an entirely non-block model) by referring to it through its item model file.
- Various NeoForge fields are not implemented (TODO: List exactly what's missing)
- `rotation` for objects in the `face` list isn't checked for. I'm referring to the field that controls UV rotation, as the `rotation`
- `rescale` is deserialized but never used
- Models can auto-register icons that are in mod directories, but can't do it from resource packs 
### Rendering
- Models are not rendered with ambient occlusion even with `ambientocclusion` set to `true`
- Block cracking is not rendered correctly
[...]

PR #373, opened 2026-05-28
Add more model patches

ah-OOG-ah:
99 little bugs in the code, 99 little bugs,
Take one down, patch it around, 127 little bugs in the code!

Fix:
- Rotated models having incorrect lighting
- Inset faces using self light for opaque blocks
- Missing model not propagating to children
- Some crashes with malformed texture packs
- Randomized model selection


Roadhog360 (approved):

Roadhog360:
Actually, some commit after the initial one for seeded random seems to have broken y value calculations. It looks like only x and z are seeded now. Weird this happens, I looked at the code and see x, y and z are all seeded.
image

Roadhog360:
Here's a better example. The same model is used 100% of the time in each column.

image

Roadhog360:
Is this possibly related to how RNG seeds work? Since each seed only has slightly different bits, and the seed changes are fairly linear along the y level, would it lead to this? If it's due to that, then I recommend adding a hasher to CoordinatePacker and using that instead  (See #243 for a hashing formula to generate a unique hash for any number)

Roadhog360:
I can't reproduce this using the rng-test block either, that's weird...

PR #377, opened 2026-05-30
Choke Thaum bore cvis source lookup (mixin)

LazyFlesh:
Add cooldown to thaum's arcane bore cvis handling to decrease calls for cvis source  checking. Should icnrease performance, especially since bores are usually spammed by hte hundreds.

If it has redstone, cooldown is 10s. If not, 30.

If it has any speedy time cached, the cooldown is skipped as long as its refilling. Otherwise, if it just has a bunch cached but isn't refilling or draining, it goes to 1/2s. This is still 10x less calls, even if they get suck in a state of only having some cvis forever.

oh, and update buildscript.


LazyFlesh:
Ill add it to hodgepodge instead of here, as directed by boubou in discord.

...tomorrow.

PR #378, opened 2026-05-30
Avoid NBT matching

AnsonYeung:
Don't match NBT if insertion is not forced and target slot is full. Some item have max stack size = 1 and checking nbt when the transfer algorithm will not overstack them is wasteful. This PR use the same function as the insertion calculation for determining whether the target do overstacking.

Before:
image
After:
image


src/main/java/com/gtnewhorizon/gtnhlib/item/StandardInventoryIterator.java
```diff
@@ -107,7 +107,7 @@ public int insert(ImmutableItemStack stack, boolean forced) {
 
         ItemStack inSlot = getStackInSlot(slotIndex);
 
-        if (!ItemUtil.isStackEmpty(inSlot) && !stack.matches(inSlot)) {
+        if (!ItemUtil.isStackEmpty(inSlot) && (!forced && !inSlot.isStackable() || !stack.matches(inSlot))) {
```
Nikolay-Sitnikov: You could probably extend this check to also include stacks that are entirely full.

```suggestion
        if (!ItemUtil.isStackEmpty(inSlot) && (!forced && (!inSlot.isStackable() || inSlot.stackSize >= inSlot.getMaxStackSize()) || !stack.matches(inSlot))) {
```

Nikolay-Sitnikov: Actually, this is incorrect, you probably want to use `getSlotStackLimit` instead.

src/main/java/com/gtnewhorizon/gtnhlib/item/StandardInventoryIterator.java
Nikolay-Sitnikov, line 110: Also, theoretically you don't need to create a partial copy of the input `ItemStack` if the stack in the slot already has that exact item (`matches`).

AnsonYeung: In theory yes, in practice this is needed to satisfy the types (ImmutableItemStack and ItemStack).

AnsonYeung:
I am not sure whether isStackable is needed, because I can't find a stack to test. Regardless I think the behaviour should be that it is needed. The difference seems to be that if there is an item that can be stacked > 1 and can be damaged, then if it's damaged it's not stackable. I think there is no such item but the code is different so I add it.

Most item that can be damaged like armor have max stack = 1, and isStackable and full stack check is the same.

Let me do some testing after minecraft auth is back up

AnsonYeung:
I've checked the code, and it looks like that for stackable item with damage (like gt meta item) they will just set maxDamage=0 which make isStackable=true. I think the item logic should respect vanilla and also check isStackable as a conclusion.

Luca-Guettinger (approved):

PR #380, opened 2026-05-30
Added color system to unify color resources

Ranzuu:
The idea is to have one global color ~enum~ class, which will unify all of them from different mods.

Later TODO:
- Add support for Alp custom formating (shadows for strings as example, colors not needed)
- Redo these mods to support this class (Might be more, I found them using different enum classes:

  - [Vending Machine](https://github.com/GTNewHorizons/VendingMachine/blob/master/src/main/java/com/cubefury/vendingmachine/util/GuiParams.java) - https://github.com/GTNewHorizons/VendingMachine/pull/138
  - [AE2](https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/blob/e099e3f319849c507855c0a7de40f28116f15358/src/main/java/appeng/core/localization/GuiColors.java) - https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/1343
     \+ class change according to https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/439 
   (https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/440)
     \+ class change according to https://github.com/GTNewHorizons/ThaumicEnergistics/pull/134
   (https://github.com/GTNewHorizons/ThaumicEnergistics/pull/135)
  - [NEI](https://github.com/GTNewHorizons/NotEnoughItems/blob/f8cec9a78c449602cbba5346e8210a09a8dd4331/src/main/java/codechicken/nei/NEIClientUtils.java#L534) - https://github.com/GTNewHorizons/NotEnoughItems/pull/940
  - [AspectRecipeIndex](https://github.com/GTNewHorizons/AspectRecipeIndex/blob/master/src/main/java/com/gtnewhorizons/aspectrecipeindex/util/Util.java) - https://github.com/GTNewHorizons/AspectRecipeIndex/pull/7
  - [Better Questing](https://github.com/GTNewHorizons/BetterQuesting/blob/3f9291958bee49841168aa7ce443ccadce95d40d/src/main/java/bq_standard/integration/nei/QuestRecipeHandler.java#L326) - https://github.com/GTNewHorizons/BetterQuesting/pull/228
  - [Nei Custom Diagram](https://github.com/GTNewHorizons/nei-custom-diagram/blob/master/src/main/java/com/github/dcysteine/neicustomdiagram/api/draw/Draw.java) - https://github.com/GTNewHorizons/nei-custom-diagram/pull/75
  - [BlockRenderer6343](https://github.com/GTNewHorizons/BlockRenderer6343/blob/927ec6392ca24b79af4651637c1a1728ba661550/src/main/java/blockrenderer6343/ClientProxy.java#L49) - https://github.com/GTNewHorizons/BlockRenderer6343/pull/50
  - [CodeChickenCore](https://github.com/GTNewHorizons/CodeChickenCore/blob/08f3e4d42a3685d2bf4794f5f3c94710a35ad0c8/src/main/java/codechicken/lib/colour/LocalizedColours.java) - https://github.com/GTNewHorizons/CodeChickenCore/pull/47
  - [mobs-info](https://github.com/GTNewHorizons/Mobs-Info/blob/4888e4c9e5bf2bba1e4c72e6893f3695f29fc7e3/src/main/java/com/kuba6000/mobsinfo/nei/EnumColors.java) - https://github.com/GTNewHorizons/Mobs-Info/pull/17
  - [tinker](https://github.com/GTNewHorizons/TinkersConstruct/blob/6cefb9b2fdf7786fad4c3c491a1d3f0f4dc11b5d/src/main/java/tconstruct/library/util/LocalizedColors.java) - https://github.com/GTNewHorizons/TinkersConstruct/pull/293


Nikolay-Sitnikov:
I am not a fan of reflective magic, but I'll let others decide on whether this is a good idea or not. I would prefer something wordy over something which only works in limited circumstances.

Also, didn't we literally have a period where we removed reflection from the way magic mods loaded their items & blocks? Why would we want to add more slow reflection magic to the start up time?

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,178 @@
...
+        cachedColor = color;
+        return color;
+    }
+
+    /** Clears the color cache on resource reload (F3+T). Registered by GTNHLib's client proxy. */
+    public static class CacheReloadListener implements IResourceManagerReloadListener {
+
+        @Override
+        public void onResourceManagerReload(IResourceManager resourceManager) {
+            synchronized (INSTANCES) {
```
Nikolay-Sitnikov, line 165: I don't think this works unless you also use `synchronized (INSTANCES)` in the constructor too, since the constructor could still modify the `INSTANCES` set while you're iterating over it.

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Returns the mod ID by checking {@link Mod} first, then asking FML which mod owns the declaring class's package.
+     * Falls back to the declaring class's simple name if neither resolves.
+     */
+    public String getModId() {
```
ah-OOG-ah: Why isn't this just a `String modid` parameter on the constructor?

ah-OOG-ah (dismissed):

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,167 @@
...
+ * 
+ * <pre>
+ * <code>
+ *  public static class MyColors {
+ *      private static final ColorResource.Factory color = new ColorResource.Factory("mymod");
+ *
+ *      public static final ColorResource
+ *      // spotless:off
+ *          background      = color.rgb("background",       "202020"),
+ *
```
Nikolay-Sitnikov: Why empty line here?

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,167 @@
...
+     * <p>
+     * Example usage:
+     * 
+     * <pre>
+     * <code>
+     *  GuiDraw.drawRect(x, y, w, h, MyColors.background.getColor());
+     * </code>
+     * </pre>
+     */
+
```
Nikolay-Sitnikov: Empty line, again.

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    private static int parseHex(String hex, boolean argb) {
+        try {
+            long value = Long.parseLong(hex.trim(), 16);
+            return argb ? (int) value : (int) (0xFF000000L | value);
+        } catch (NumberFormatException e) {
+            GTNHLib.LOG.warn("[ColorResource] Invalid hex '{}', using opaque white.", hex);
+            return 0xFFFFFFFF;
```
Nikolay-Sitnikov: Don't do this. This (default) color is specified directly by the developer. If the developer made a mistake, they should get a crash, not a quiet warning and poorly-functioning code.

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,167 @@
...
+     * 
+     * <pre>
+     * <code>
+     *  GuiDraw.drawRect(x, y, w, h, MyColors.background.getColor());
+     * </code>
+     * </pre>
+     */
+
+    public int getColor() {
+        if (cachedColor != -1L) return (int) cachedColor;
```
Nikolay-Sitnikov: Technically speaking, this is wrong. If `CacheReloadListener` sets the color to -1 between the two field gets, you could accidentally return -1. Remember, if the field is volatile, *every single usage* of that field is its own independent volatile read/write.

You'd need to put the color into a local variable first, and then run the check using the local variable.

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,167 @@
...
+            try {
+                long parsed = Long.parseLong(value, 16);
+                color = argb ? (int) parsed : (int) (0xFF000000L | parsed);
+            } catch (NumberFormatException e) {
+                GTNHLib.LOG.warn("[ColorResource] Invalid hex '{}' for lang key '{}', using default.", value, langKey);
+                color = defaultColor;
+            }
+        } else {
+            color = defaultColor;
+        }
```
Nikolay-Sitnikov: Technically speaking, if two threads try to get the color at the same time, both threads could try to calculate the color at the same time. This is likely not an issue (other than the potential double log), but you could use the [double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking) pattern here to make this code more correct.

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,167 @@
...
+
+    public int getColor() {
+        if (cachedColor != -1L) return (int) cachedColor;
+
+        int color;
+        if (StatCollector.canTranslate(langKey)) {
+            String value = StatCollector.translateToLocal(langKey).trim();
+            try {
+                long parsed = Long.parseLong(value, 16);
+                color = argb ? (int) parsed : (int) (0xFF000000L | parsed);
```
Nikolay-Sitnikov: Do you want to add a warning in case someone tries to set a ARGB value for a RGB resource?

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /** Clears the color cache on resource reload (F3+T). Registered by GTNHLib's client proxy. */
+    public static class CacheReloadListener implements IResourceManagerReloadListener {
+
+        @Override
+        public void onResourceManagerReload(IResourceManager resourceManager) {
+            synchronized (INSTANCES) {
+                for (ColorResource instance : INSTANCES) {
+                    instance.cachedColor = -1L;
```
Nikolay-Sitnikov: The vanilla language manager reload listener is (most likely, unless we do some future multithreading optimizations) going to run before yours, so you could just compute the color values in this method and skip all that lazy-loading logic.

Nikolay-Sitnikov (dismissed):

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
Ranzuu: To distinguish argb vs rgb more easly
image

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
Ranzuu: removed

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
Ranzuu: So pretty much do I remove that try and catch error part?

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
Ranzuu: Alright. Local one need to be `long` right?

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
Ranzuu: That also will need to use the local (but assume without the `long`, because its already declared)?
```java
cached = cachedColor;
	 if (cached != -1L) return (int) cached;
```

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
Ranzuu: Would be nice, so like `value.length() > 6` and like Is NOT ARGB?

`!argb && value.length() > 6`

src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
Ranzuu: So I guess you can discard my reply comments, because that will make them remove if I understood this correctly.
so we dont do another check, and just directly return the `cachedColor`? Which needs to be integer now also?

iouter:
I suggest using custom JSON to store colors instead of lang. the values in lang files may be replaced by translations.
related issue: https://github.com/GTNewHorizons/GTNH-Translations/issues/50

Ranzuu:
> I suggest using custom JSON to store colors instead of lang. the values in lang files may be replaced by translations. related issue: [GTNewHorizons/GTNH-Translations#50](https://github.com/GTNewHorizons/GTNH-Translations/issues/50)

This is optional localization for resource packs, these key are not present in the lang file by default

As for the ae2 `gui.colors` inside lang, they can be deleted from there. Same stuff - optional for override. I will fix it with the change to this system.

iouter:
> > I suggest using custom JSON to store colors instead of lang. the values in lang files may be replaced by translations. related issue: [GTNewHorizons/GTNH-Translations#50](https://github.com/GTNewHorizons/GTNH-Translations/issues/50)
> 
> This is optional localization for resource packs, these key are not present in the lang file by default
> 
> As for the ae2 `gui.colors` inside lang, they can be deleted from there. Same stuff - optional for override. I will fix it with the change to this system.

So does that mean we need to delete them from en_US.lang? After all, as long as they are in en_US.lang, they will be uploaded to paratranz and then translated.

Ranzuu:
> > > I suggest using custom JSON to store colors instead of lang. the values in lang files may be replaced by translations. related issue: [GTNewHorizons/GTNH-Translations#50](https://github.com/GTNewHorizons/GTNH-Translations/issues/50)
> > 
> > 
> > This is optional localization for resource packs, these key are not present in the lang file by default
> > As for the ae2 `gui.colors` inside lang, they can be deleted from there. Same stuff - optional for override. I will fix it with the change to this system.
> 
> So does that mean we need to delete them from en_US.lang? After all, as long as they are in en_US.lang, they will be uploaded to paratranz and then translated.

You don't need them in the lang file, so yes you can delete them. Unless you want different colors (which I assume is a nono)

cubefury (approved):
looking good, thanks for the effort with the unification

PR #385, opened 2026-06-02
Best effort avoidance of extraction

AnsonYeung:
An imperfect solution for https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/24956

The issue is the underlying ItemTransfer API itself and cannot be completely fixed.

If the extract is done on an ae2 interface connected to an ae2 network that only have extract only storage bus, inserting back into the ae2 storage is just not possible. The default implementation of ae2 interface always queries real network storage on extract.

Also, we should try our best to avoid any ae2 queries. AE2 query operations is one of the slowest thing on the server and must be avoided if we can. The creation of many item movement even when the target inventory is full is not very acceptable.

The implementation is best-effort. There is no guarantee that this completely fixes the issue but it should be enough for the common cases.


AnsonYeung:
I mark this as draft, maybe should be closed but I think the ae2 interface fix is great. This is not really needed now.

AnsonYeung:
actually I'll just close it

PR #391, opened 2026-06-07
Fix dep manifest crash on fplib older than 1.11

alppp:
fplib before 1.11 streams the always/obf/dev scopes without null-checking, so manifests that only declare `always` NPE in scanDeps. Declaring obf/dev empty fixes it and is a no-op on 1.11+. Hits any mod bumping GTNHLib to >=0.11 under an old resolved fplib (floor is still 1.9.0)


ah-OOG-ah:
According to Alp, this was fixed elsewhere

PR #394, opened 2026-06-08
Fix deps for mods depending on mods using jvmdg

mitchej123:
Fix dependencies so mods that don't use jvmdg but depend on on mods that use jvmdg work.  Also play nicer with fplib deploader in dev


koolkrafter5 (approved):
We love random problems that come up right before a major release

mitchej123:
To be fair, this changes nothing about runtime (minus the fastutil bump), it's only for dev ;)

PR #396, opened 2026-06-10
wrap ChatComponentItemName text in brackets like vanilla func_151000_E()

Eldrinn-Elantey:
getUnformattedTextForChat() returned just the item name without brackets, unlike vanilla func_151000_E() which wraps it in [brackets]. This made the chat component indistinguishable from plain text. Wrapped in brackets so it matches vanilla format and can be reliably identified in chat.


MalTeeez (approved):
LGTM, but screenshots might not hurt in the future

PR #399, opened 2026-06-11
add get keybind name

lc-1337:


Kogepan229:
It seems GameSettings.getKeyDisplayString is in vanilla, so this PR might be unnecessary.

PR #400, opened 2026-06-13
Update docs on ColorResource

Ranzuu:


Algent (approved):
It's my fault isn't it lmao

PR #401, opened 2026-06-14
Optimize ColorResource class

Ranzuu:
Currently the class is loading all lang keys and waits for the resource pack to "show them". This is super wrong, especially with some example here: https://github.com/GTNewHorizons/GT5-Unofficial/pull/7144

`langKey` was removed as a stored field instead `getLangKey()` builds it on demand. No 1515 (example, number of entries) permanent strings sitting in heap. Memory drops from ~150 KB (just the strings) to essentially zero between reloads.

`ALL` + `ACTIVE` instead of `INSTANCES` - `ALL` holds every instance (same as before). `ACTIVE` is a plain `ArrayList` holding only instances that have a confirmed resource pack override.

- First reload (game start, resource pack loaded) -> Scan all 1515 -> build ACTIVE
- `F3+T`, no resource pack -> `ACTIVE` is empty; Zero lookups
- `F3+T`, pack with overrides -> A`CTIVE` has X entries; X lookups (N is small)
- Override disappears (resource pack unloaded) -> Entry removed from `ACTIVE`; reset to default

`invalidate()` resets the flag, to force a full rescan on the next reload 

(Would matter is if a user opens the resource pack GUI mid-game, swaps a pack, and the new pack adds color overrides)

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
@@ -123,8 +123,9 @@ public class ColorResource {
 
     private int resolveColor() {
         String langKey = getLangKey();
-        if (StatCollector.canTranslate(langKey)) {
-            String value = stripPrefix(StatCollector.translateToLocal(langKey));
+        String translatedKey = StatCollector.translateToLocal(langKey);
+        if (langKey != translatedKey) {
+            String value = stripPrefix(translatedKey);
             try {
                 if (!argb && value.length() > 6) {
                     LOG.warn(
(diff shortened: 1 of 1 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/color/ColorResource.java
```diff
@@ -110,6 +122,7 @@ public int getColor() {
     }
 
     private int resolveColor() {
+        String langKey = getLangKey();
         if (StatCollector.canTranslate(langKey)) {
             String value = stripPrefix(StatCollector.translateToLocal(langKey));
```
sisyphussy: You can get rid of the `canTranslate` by just calling `translateToLocal` and checking if `langKey != translatedKey`

Ranzuu: oh that would change the lookup from hash map from 2 to 1? (so, good)

sisyphussy: Yes, should make it a lot faster to load.

Nikolay-Sitnikov (approved):

PR #407, opened 2026-06-26
Update team commands

cubefury:
Makes it so the suggested command to accept a team invite/merge will only show the team name if there's more than one active invite. Otherwise it'll just show the base accept/reject command.

image


MassAnarchyy (approved):
Nice :)

PR #409, opened 2026-06-27
Add world conversion warning system

SuperSoupr:
This is a simple system for showing custom warning screens when trying to load a world where items and blocks need to be converted from one mod to another

image

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/api/gui/IWorldConversionWarning.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/api/gui/IWorldConversionWarning.java
@@ -41,5 +41,7 @@ public interface IWorldConversionWarning {
      * @return Gui to be used in the warning. null to use default.
      */
     @SideOnly(Side.CLIENT)
-    GuiConfirmationWCW getGui(StartupQuery startupQuery);
+    default GuiConfirmationWCW getGui(StartupQuery startupQuery) {
+        return null;
+    }
 }
--- a/src/main/java/com/gtnewhorizon/gtnhlib/api/gui/IWorldConversionWarning.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/api/gui/IWorldConversionWarning.java
@@ -4,7 +4,7 @@ import cpw.mods.fml.common.StartupQuery;
 import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
 
-public abstract class WorldConversionWarning {
+public interface IWorldConversionWarning {
 
     /**
      * Will be called right after {@link cpw.mods.fml.common.event.FMLMissingMappingsEvent}s are fired to determine if
@@ -12,14 +12,14 @@ public abstract class WorldConversionWarning {
      *
      * @return True if the warning should be shown
      */
-    public abstract boolean shouldShow();
+    boolean shouldShow();
 
     /**
      * Server-side only a simple text message will be printed.
      *
      * @return The text message
      */
-    public abstract String getServerMessage();
+    String getServerMessage();
 
     /**
      * Client-side you can choose to either supply a custom {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW}
@@ -29,7 +29,7 @@ public abstract class WorldConversionWarning {
      * @return The text message for default gui class
      */
     @SideOnly(Side.CLIENT)
-    public abstract String getClientMessage();
+    String getClientMessage();
 
     /**
      * Children MUST annotate with @SideOnly(Side.CLIENT)
@@ -41,5 +41,5 @@ public abstract class WorldConversionWarning {
      * @return Gui to be used in the warning. null to use default.
      */
     @SideOnly(Side.CLIENT)
-    public abstract GuiConfirmationWCW getGui(StartupQuery startupQuery);
+    GuiConfirmationWCW getGui(StartupQuery startupQuery);
 }
--- a/src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarning.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarning.java
@@ -1,0 +1,45 @@
+package com.gtnewhorizon.gtnhlib.api.gui;
+
+import cpw.mods.fml.common.StartupQuery;
+import cpw.mods.fml.relauncher.Side;
+import cpw.mods.fml.relauncher.SideOnly;
+
+public abstract class WorldConversionWarning {
+
+    /**
+     * Will be called right after {@link cpw.mods.fml.common.event.FMLMissingMappingsEvent}s are fired to determine if
+     * the warning should be shown.
+     *
+     * @return True if the warning should be shown
+     */
+    public abstract boolean shouldShow();
+
+    /**
+     * Server-side only a simple text message will be printed.
+     *
+     * @return The text message
+     */
+    public abstract String getServerMessage();
+
+    /**
+     * Client-side you can choose to either supply a custom {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW}
+     * or the default one with a text message. To show a default
+     * {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW} return null in {@link #getGui(StartupQuery)}.
+     *
+     * @return The text message for default gui class
+     */
+    @SideOnly(Side.CLIENT)
+    public abstract String getClientMessage();
+
+    /**
+     * Children MUST annotate with @SideOnly(Side.CLIENT)
+     * <p>
+     * Client-side you can choose to either supply a custom {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW}
+     * or the default one with a text message. To show a default
+     * {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW} return null.
+     *
+     * @return Gui to be used in the warning. null to use default.
+     */
+    @SideOnly(Side.CLIENT)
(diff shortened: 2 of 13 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/mixins/preinit/MixinGameData_WorldConversionWarning.java
```diff
@@ -0,0 +1,44 @@
...
+    private static void uie$processIdRematches(Iterable<FMLMissingMappingsEvent.MissingMapping> missedMappings,
+            boolean isLocalWorld, GameData gameData, Map<String, Integer[]> remaps,
+            CallbackInfoReturnable<List<String>> cir) {
+        WorldConversionWarningManager.WARNINGS.forEach((id, wcw) -> {
+            if (wcw.shouldShow()) {
+                boolean confirmed = StartupQuery
+                        .confirm(FMLCommonHandler.instance().getSide().isServer() ? wcw.getServerMessage() : id);
+                if (!confirmed) StartupQuery.abort();
+            }
+        });
```
C0bra5: (optional) Might be worth moving out to a util class to make dev-env debugging easier.

src/main/java/com/gtnewhorizon/gtnhlib/mixins/preinit/MixinGameData_WorldConversionWarning.java
```diff
@@ -0,0 +1,44 @@
...
+    // This pre-init mixin is required to inject into the exact moment after missing mapping events are fired
+    // but before the confirmation message appears.
+    @Definition(id = "defaulted", local = @Local(type = List.class, name = "defaulted"))
+    @Definition(id = "isEmpty", method = "Ljava/util/List;isEmpty()Z")
+    @Expression("defaulted.isEmpty()")
+    @Inject(
+            method = "processIdRematches",
+            at = @At(value = "MIXINEXTRAS:EXPRESSION", shift = At.Shift.BY, by = -2),
+            remap = false)
+    private static void uie$processIdRematches(Iterable<FMLMissingMappingsEvent.MissingMapping> missedMappings,
```
C0bra5: Wrong prefix used
```suggestion
    private static void gtnhlib$processIdRematches(Iterable<FMLMissingMappingsEvent.MissingMapping> missedMappings,
```

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/IWorldConversionWarning.java
C0bra5, line 1: Rename to IWorldConversionWarning.java

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarning.java
```diff
@@ -0,0 +1,45 @@
+package com.gtnewhorizon.gtnhlib.api.gui;
+
+import cpw.mods.fml.common.StartupQuery;
+import cpw.mods.fml.relauncher.Side;
+import cpw.mods.fml.relauncher.SideOnly;
+
+public abstract class WorldConversionWarning {
```
C0bra5: This should be an interface
```suggestion
public interface IWorldConversionWarning {
```

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarningManager.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Allows to register more confirmation screens to show before the FML missing mapping warning screen. Specifically
+ * designed for world conversion warnings, these will only fire if the currently loading world contains any missing
+ * mapping. Fires *after* {@link cpw.mods.fml.common.event.FMLMissingMappingsEvent}s but *before* the warning gui /
+ * server message.
+ */
+public class WorldConversionWarningManager {
+
+    public static final HashMap<String, WorldConversionWarning> WARNINGS = new HashMap<>();
```
C0bra5:
```suggestion
    public static final HashMap<String, IWorldConversionWarning> WARNINGS = new HashMap<>();
```

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarningManager.java
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * Call during game startup to register a {@link WorldConversionWarning}.
+     *
+     * @param id A unique id for your warning
+     */
+    public static void register(String id, WorldConversionWarning wcw) {
```
C0bra5:
```suggestion
    public static void register(String id, IWorldConversionWarning wcw) {
```

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarning.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Will be called right after {@link cpw.mods.fml.common.event.FMLMissingMappingsEvent}s are fired to determine if
+     * the warning should be shown.
+     *
+     * @return True if the warning should be shown
+     */
+    public abstract boolean shouldShow();
```
C0bra5: Might be worth adding a default interface implementation that returns true.
```suggestion
    boolean shouldShow();
```

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarning.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Server-side only a simple text message will be printed.
+     *
+     * @return The text message
+     */
+    public abstract String getServerMessage();
```
C0bra5:
```suggestion
    String getServerMessage();
```

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarning.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Client-side you can choose to either supply a custom {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW}
+     * or the default one with a text message. To show a default
+     * {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW} return null in {@link #getGui(StartupQuery)}.
+     *
+     * @return The text message for default gui class
+     */
+    @SideOnly(Side.CLIENT)
+    public abstract String getClientMessage();
```
C0bra5:
```suggestion
    String getClientMessage();
```

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarning.java
```diff
@@ -0,0 +1,45 @@
...
+     * Children MUST annotate with @SideOnly(Side.CLIENT)
+     * <p>
+     * Client-side you can choose to either supply a custom {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW}
+     * or the default one with a text message. To show a default
+     * {@link com.gtnewhorizon.gtnhlib.api.gui.GuiConfirmationWCW} return null.
+     *
+     * @return Gui to be used in the warning. null to use default.
+     */
+    @SideOnly(Side.CLIENT)
+    public abstract GuiConfirmationWCW getGui(StartupQuery startupQuery);
```
C0bra5: Might be worth implementing a default implementation with the default class.
```suggestion
    GuiConfirmationWCW getGui(StartupQuery startupQuery);
```

C0bra5 (dismissed):
IDK if GTNHLib already includes something for this, but it would be nice if you could expand the scope of this system to be able to add per-load warnings that aren't tied to world conversions for things like:
- Notifying players/devs of added mods (useful for detecting transitive dep changes that you might not have expected as a dev)
- Easily notifying players of breaking changes.

src/main/java/com/gtnewhorizon/gtnhlib/api/gui/WorldConversionWarning.java
SuperSoupr: Oops, there used to be a reason for it being an abstract class but I forgot I could change it back to interface

SuperSoupr:
> IDK if GTNHLib already includes something for this, but it would be nice if you could expand the scope of this system to be able to add per-load warnings that aren't tied to world conversions for things like:
> 
>     * Notifying players/devs of added mods (useful for detecting transitive dep changes that you might not have expected as a dev)
> 
>     * Easily notifying players of breaking changes.

I don't think this is the right place for that, these are fired like the FML missing mappings' screen, so under the hood they lock the client and server threads in a while loop with a `Thread.sleep(200)` call in it. What you're describing would be much better if implemented by redirecting the load world button click and reading level.dat

Also, while I personally like the idea of a breaking changes page, it's sure to turn into big discussion which I don't want to bring to this pr

C0bra5:
> > IDK if GTNHLib already includes something for this, but it would be nice if you could expand the scope of this system to be able to add per-load warnings that aren't tied to world conversions for things like:
> > ```
> > * Notifying players/devs of added mods (useful for detecting transitive dep changes that you might not have expected as a dev)
> > 
> > * Easily notifying players of breaking changes.
> > ```
> 
> I don't think this is the right place for that, these are fired like the FML missing mappings' screen, so under the hood they lock the client and server threads in a while loop with a `Thread.sleep(200)` call in it. What you're describing would be much better if implemented by redirecting the load world button click and reading level.dat
> 
> Also, while I personally like the idea of a breaking changes page, it's sure to turn into big discussion which I don't want to bring to this pr

Fair enough, it's why I left it in the comment as a suggestion since it would be nice to have in the future, and would probably share some of the groundwork.

C0bra5 (approved):

PR #418, opened 2026-07-08
Fix NumberPattern

Sanduhr32:
## Summary
Fixes the regex for numbers, it was inprecisely capturing a lot of invalid notations for 8-64bit integers & captured integers above 64bit.

Tested values:
000000
-0000000
-11111
111111
1.
-2.9
2.9
+2.9
1e1
1.e1
1e-1
-1e8
+8e4
-2374623.233525e+23423

invalid values:
1e   missing at least 1 digit for exponent
digits of integers as strings longer than 20 chars

## Checklist

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
@@ -28,7 +28,7 @@ import com.google.gson.JsonPrimitive;
 // Taken from NEI
 public class NBTJson {
 
-    private static final Pattern numberPattern = Pattern.compile("^([-+]?\\d+\\.?\\d*E*\\d*)([bBsSlLfFdD]?)$");
+    private static final Pattern numberPattern = Pattern.compile("^([-+]?\\d+\\.?\\d*E*\\d{,20)([bBsSlLfFdD]?)$");
 
     public static String toJson(NBTTagCompound tag) {
         return toJson(toJsonObject(tag));
@@ -153,7 +153,8 @@ public class NBTJson {
                     }
                 } else {
                     if (numberString.contains(".")) return new NBTTagDouble(Double.parseDouble(numberString));
-                    else return new NBTTagInt(Integer.parseInt(numberString));
+                    else if (numberString.length() <= 12) return new NBTTagInt(Integer.parseInt(numberString));
+                    else return new NBTTagString(jsonString);
                 }
             } else {
                 // String
--- a/src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
@@ -32,7 +32,7 @@ public class NBTJson {
             .compile("^([-+]?\\d{1,20}(?:\\.?\\d*(?:[eE][-+]?\\d+)?)?)([bBsSlLfFdD]?)$");
     /**
      * in case someone actually puts a "+" signed unsigned int into a json
-     * 
+     *
      * @see Integer#parseUnsignedInt(String)
      */
     private static final int UNSIGNED_INT_STRING_LENGTH = 12;
--- a/src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
@@ -28,7 +28,7 @@ import com.google.gson.JsonPrimitive;
 // Taken from NEI
 public class NBTJson {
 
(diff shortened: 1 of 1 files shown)


Ruling-0 (approved):
Looks good, thanks!

src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
```diff
@@ -28,7 +28,14 @@
 // Taken from NEI
 public class NBTJson {
 
-    private static final Pattern numberPattern = Pattern.compile("^([-+]?\\d+\\.?\\d*E*\\d*)([bBsSlLfFdD]?)$");
+    private static final Pattern numberPattern = Pattern
+            .compile("^([-+]?\\d{1,20}(?:\\.?\\d*(?:[eE][-+]?\\d+)?)?)([bBsSlLfFdD]?)$");
```
MalTeeez: You are allowing trailing dots here, I'd turn `\.?\d*` into `\.?\d+`

src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
MalTeeez: While you are here, change this to .matches()

src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
```diff
@@ -153,7 +160,9 @@ public static NBTBase toNbt(JsonElement jsonElement) {
                     }
                 } else {
                     if (numberString.contains(".")) return new NBTTagDouble(Double.parseDouble(numberString));
-                    else return new NBTTagInt(Integer.parseInt(numberString));
+                    else if (numberString.length() <= UNSIGNED_INT_STRING_LENGTH)
+                        return new NBTTagInt(Integer.parseInt(numberString));
+                    else return new NBTTagString(jsonString);
```
MalTeeez, line 162: I wouldn't depend on the string length for digit length here, just do: 
```suggestion
                    try {
                        return new NBTTagInt(Integer.parseInt(numberString));
                    } catch (NumberFormatException ignored) {
                        return new NBTTagString(jsonString);
                    }
```

MalTeeez: (Don't forget to import that exception)

src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
```diff
@@ -153,7 +160,9 @@ public static NBTBase toNbt(JsonElement jsonElement) {
                     }
                 } else {
                     if (numberString.contains(".")) return new NBTTagDouble(Double.parseDouble(numberString));
```
MalTeeez: This won't catch exponent notation which will fail at the Integer construction. Since Double should accept that, id also guard for that here:
```suggestion
                    if (numberString.indexOf('.') >= 0
                            || numberString.indexOf('e') >= 0
                            || numberString.indexOf('E') >= 0)
                        return new NBTTagDouble(Double.parseDouble(numberString));
```

MalTeeez (commented):
some edge cases

src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
Sanduhr32: Double parsing allows it, i'm no fan of it either

src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
Sanduhr32: exceptions are way slower here, and i can guarantee that each team that starts to use wireless energy once they can will always cause an exception to be thrown & caught here when the team gets loaded.
It also doesnt require extra allocations, neither does it attempt to parse ints that arent flagged as longs or big ints (wireless energy)

MalTeeez: Sounds good, as long as we don't let the exponents slip through above 👍

src/main/java/com/gtnewhorizon/gtnhlib/util/NBTJson.java
MalTeeez: Afaik double parsing also allows stuff like .5 which isn't great either - id enforce at least some sanity here

Sanduhr32: it d+ now, so 1 or any digits

MalTeeez (approved):
Awesome, ty!

PR #423, opened 2026-07-12
Add bukkit compatibility Layer

### LINKED ISSUE #25797: gtnhteam command doesn't work on Thermos

Live_Learn

### Your Pack Version

2.9.0 beta 2

### Your Server

Zeta

### Java Version

Java 25

### Type of Server

Thermos

### Your Expectation

Trying to use `/gtnhteam` command to team with other players on Zeta

### The Reality

Unknown command. Type "/help" for help.

Image

### Your Proposal

Fix it. Maybe it's related to https://github.com/GTNewHorizons/GTNHLib/issues/411 ?

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

RealSilverMoon:
## Summary
Brigadier is incompatible with bukkit command registry. None of the commands registered by brigadierAPI would work on Thermos.
This compatibility Layer will register brigadier commands as standard commands **only on thermos server**.
Close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/25797
Tested on a latest Crucible core server with daily 597.

## Checklist
- [x] I have tested this PR in Fullpack

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
@@ -130,8 +130,20 @@ public class CommonProxy {
         event.registerServerCommand(new BlockStateCommand());
         event.registerServerCommand(new TitleCommand());
 
-        for (CommandNode<ICommandSender> node : BrigadierApi.getCommandDispatcher().getRoot().getChildren()) {
-            event.registerServerCommand(new BrigadierCommandWrapper(node.getName()));
+        if (isThermosServer()) {
+            for (CommandNode<ICommandSender> node : BrigadierApi.getCommandDispatcher().getRoot().getChildren()) {
+                event.registerServerCommand(new BrigadierCommandWrapper(node.getName()));
+            }
+        }
+    }
+
+    public static boolean isThermosServer() {
+        try {
+            Class.forName("thermos.ThermosRemapper");
+            GTNHLib.LOG.warn("Thermos detected, applying command wrapper");
+            return true;
+        } catch (ClassNotFoundException e) {
+            return false;
         }
     }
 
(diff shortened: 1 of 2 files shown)


spacebuilder2020:
Does this work with other bukkit variants such as crucible or does it only work with thermos.  I ask because from a glance it looks like it is checking for a thermos specific class.

FE does a similar check but that check is old so it may not be as effective either. https://github.com/ForgeEssentials/ForgeEssentials/blob/1.7.10/develop/src/main/java/com/forgeessentials/core/preloader/FELaunchHandler.java#L84

Edit: disregard, I see you mentioned crucible in your description.

src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
```diff
@@ -126,6 +129,22 @@ public void serverAboutToStart(FMLServerAboutToStartEvent event) {
...
+        if (isThermosServer()) {
+            for (CommandNode<ICommandSender> node : BrigadierApi.getCommandDispatcher().getRoot().getChildren()) {
+                event.registerServerCommand(new BrigadierCommandWrapper(node.getName()));
+            }
+        }
+    }
+
+    public static boolean isThermosServer() {
+        try {
+            Class.forName("thermos.ThermosRemapper");
```
spacebuilder2020: This looks fine class wise, but you might need to adjust it so it doesn't load the class when the check is run since that could potentially cause issues.  I recommend doing something like this:
`Class.forName("thermos.ThermosRemapper", false, this.getClass().getClassLoader());`

spacebuilder2020 (changes requested):
After a second review, I see you already tested it on Crucible, but there is one more thing that needs to be checked.  Remember, calling class forName by default will initialize the class if it is not already loaded.  This might be undesired for the check especially if it happens early.

src/main/java/com/gtnewhorizon/gtnhlib/CommonProxy.java
RealSilverMoon: Make sense

spacebuilder2020 (approved):
Looks Good!

PR #425, opened 2026-07-13
BlockState API + functionality improvements

RecursivePineapple:
## Summary
The BlockState system had several major design flaws that this PR intends to fix. Primarily, it required a specific BlockProperty reference to store a value in a BlockState. Since properties can't easily be arbitrarily fetched, this made deserialization essentially impossible. To fix this, property values were given various states within a BlockState. They can be 'normal,' which means the property has been validated and is known to be correct, 'unvalidated,' which means that code has injected a potentially invalid property into the state, and 'deferred,' which means that the value is stored as a string and must be resolved at runtime. For the most part, these states are handled transparently, but some BlockState get-value methods have to return a String for deferred properties. The exact details should be fully documented within the BlockState class/method documentation.

I also made pooling opt-in, to fix thread safety issues. Consumers of the API can supply their own BlockStatePool that provides BlockState references, but this is optional. There are no pool singletons to avoid thread safety issues, consumers must create their own.

Many BlockProperty implementations have new non-boxing versions of the getValue/setValue methods, for efficiency. This should be 100% backwards compatible, though implementing code will need some changes to use these new methods properly. Some interfaces were renamed, but I tried to ensure API & ABI backwards compatibility to prevent breakage.

## Checklist


ah-OOG-ah (approved):
Seems to work in beta1, and it doesn't brick UIE (current user of models)

PR #427, opened 2026-07-14
Support bake item sprite

ysesiq:
## Summary
When `JSONModel#bakeSprite` bakes a block's atlas sprite, if the `IIcon` is missingno, it will fall back to matching from `TextureMap.locationItemsTexture`.

## Checklist

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/client/model/unbaked/JSONModel.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/client/model/unbaked/JSONModel.java
@@ -253,7 +253,8 @@ public class JSONModel implements UnbakedModel {
         name = name.replaceFirst("^minecraft:", "");
         var icon = Minecraft.getMinecraft().getTextureMapBlocks().getAtlasSprite(name);
         if ("missingno".equals(icon.getIconName())) {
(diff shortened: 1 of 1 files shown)


ah-OOG-ah (changes requested):
This doesn't work. Just using the IIcon from the item atlas won't give you item textures, because only the block atlas is available when the model is rendered.

ysesiq:
> This doesn't work. Just using the IIcon from the item atlas won't give you item textures, because only the block atlas is available when the model is rendered.

You can use the check if the icon name is `missingno` to switch between binding to block atlas and item atlas. 
[this example will successfully render the item](https://github.com/ysesiq/retroEMI-Forge/blob/1.7.10/src/main/java/dev/emi/emi/api/stack/TagEmiIngredient.java#L127)

ah-OOG-ah:
> You can use the check if the icon name is `missingno` to switch between binding to block atlas and item atlas. [this example will successfully render the item](https://github.com/ysesiq/retroEMI-Forge/blob/1.7.10/src/main/java/dev/emi/emi/api/stack/TagEmiIngredient.java#L127)

That only works for items or block items, not blocks in world... which are the things JSON models are intended to render.

This PR is superceded by #438, which *does* work in-world.

PR #428, opened 2026-07-15
Add simplex noise utilities

RecursivePineapple:
## Summary
This adds some useful noise generation utils. I've used these in several projects so I figured I should add them to gtnhlib.

## Checklist


src/main/java/com/gtnewhorizon/gtnhlib/noise/ScaledSampler.java
```diff
@@ -0,0 +1,32 @@
+package com.gtnewhorizon.gtnhlib.noise;
+
+/// Scales another sampler by a certain amount in each axis.
+/// Effects are the opposite of what you'd expect - scaling by 2 in an axis shrinks the noise by half along that axis.
```
ah-OOG-ah, line 4: This implies the scale in question is affecting the numerical scale (i.e. scale 2 produces results in [-0.5, 0.5]). Might want to clarify that you're talking about coordinate scaling.

RecursivePineapple: I finally got back to this PR. The comment should be more explicit that the transform is only for the coordinates now.

erialsan:
Since merging into EFR Upstream causes the build to fail, please make a decision as soon as possible.
If the merge does not go through, other methods will be pursued, so an early decision is desirable.

ah-OOG-ah (approved):

PR #434, opened 2026-07-29
Fix walk/fall particle icons when walking on modeled blocks

ah-OOG-ah:
## Summary
Previously, modeled blocks would return missingno for the walk and fall particles. This replaces the spawning method so the full blockstate can be wired in.

## Checklist
- [ ] ~~This PR requires another PR in order to merge~~ no it doesn't


ah-OOG-ah:
note: still missing fall particles

ah-OOG-ah:
note: no longer missing fall particles

FourIsTheNumber (approved):

PR #436, opened 2026-07-31
Command to drop all assets into new resource pack

Ranzuu:
## Summary
Alright, so the idea is to have a command in-game to drop all game assets. That includes all mods that are in.

First of all, I cleanded all the resource pack class file names, so they are named correctly. It was Update Checker at first. Now it's global resourcepack utils.

Rest was AI.
- The command is disabled by default in the cofing.
- The command has confirmation alert because in fullpack this takes around 1min 20s as of the latest daily (653)

This pretty much does what I described, so just dumps all assets into the resource pack, with bunch of logs and a timestamp name. All of that is made into folder, and not zip, because god damn no. And with folder you can edit them nicely or do what you want

## Checklist
- [ ] This PR is in compliance with the [GTNH AI Policy](https://github.com/GTNewHorizons/GTNH-Dev-Doc/blob/master/AI_POLICY.md)


glowredman:
I don't see why this needs to he part of any mod and not a standalone tool (those already exist btw).
Besides, as long as this PR does not follow our AI Policy, it can't be merged. For this reason alone I'm going to close the PR, the above point could be discussed.

PR #447, opened 2026-08-11
Vanilla Color Code Rules Option + Minor JSON Crash Fix 

KAMKEEL:
## Summary

In order for the Hex Text Compatibility PR to merge for Angelica: https://github.com/GTNewHorizons/Angelica/pull/1972 

This system needs to be in place for GTNHLib so that Angelica can follow formatting rules of Hex Text as an optional config to put create a parity with Vanilla MC Behavior in which Hex Colors reset Formatting codes like Bold. If this isn't in, then string lengths are off for GTNHLib and Hex Text when joined together. This change is simply adding an API Extension for a minor point for Hex Text to modify / resolve

As for the JSON Issue. I found that some legacy JSON formats produced a crash with GTNHLib in Resource packs when loading up, this simple change just allows it support "normal" blocks formatting which some ResourcePacks carry.

## Checklist

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/client/model/loading/ResourceLoc.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/client/model/loading/ResourceLoc.java
@@ -15,6 +15,7 @@ import net.minecraft.util.ResourceLocation;
 import org.jetbrains.annotations.NotNull;
 
 import com.google.gson.Gson;
+import com.google.gson.JsonParseException;
 import com.gtnewhorizon.gtnhlib.client.model.state.StateModelMap;
 import com.gtnewhorizon.gtnhlib.client.model.unbaked.JSONModel;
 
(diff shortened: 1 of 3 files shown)


alppp:
one thing to fix, here or in a follow-up: 
- the new `hexColorResetsStyles`only applies inside the branch gated on `preprocessorHandlesAmpCodes()` in `getStringWidth` / `sizeStringToWidth` / `trimStringToWidth`. when a hextext server blocks universal ampersand, `handlesAmpCodes()` returns false, so the width code reads the `§x` payload as six legacy color pairs, drops bold, and never asks the flag. the renderer does ask it. 
- with vanillaReset off that is the bold+hex width overflow again (tooltips too wide, wrong wrap points). suggestion: gate the `§x` skip on "a preprocessor is registered" instead of `handlesAmpCodes()`, and apply `hexColorResetsStyles` on that path too.

RecursivePineapple (approved):

PR #452, opened 2026-08-18
Fix resource packs not loading at startup

### LINKED ISSUE #450: Resourcepacks resetting on load when updating from 0.11.37 to -39

_No response_

### Mod Version

gtnhlib-0.11.39

### Bug Report

Reloading them after _works_, but it still resets them on the next load.
[fml-client-latest.log](https://github.com/user-attachments/files/31127075/fml-client-latest.log)

### Java Version

Java 8

In the fml log.

### Final Checklist

sunshinekitsune:
## Summary
Fixes #450
(Related: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/26364)

In 0.11.38 a bug was introduced in nhlib$blockOldModels. The intention was to prevent legacy resource packs from loading blockstates/models however it introduced a regression where enabled resource packs weren't applied when launching the game.

Originally what would happen is that in the ResourcePackRepository constructor it would scan the resource packs folder and call getPackImage() for pack.png. nhlib$blockOldModels would intercept "pack.png" and call rprMetadataSerializer. Because of ResourcePackRepository still running the other callback would return null which would throw a unhandled NullPointerException. Minecraft would then catch the exception and drop the pack. This leads to every pack being cleared since all enabled packs would not be found in the repository. This is also why after the game loads you can reapply them without issue.

In this PR I fixed this by parsing pack.mcmeta with GSON instead of ResourcePackRepository. If I'm not mistaken, pack.mcmeta must always be valid JSON. This allows for reading the json in a slef contained way that doesn't have any lifecycle issues.

Additionally, I reordered checks for checking blockstate resources so that non-jsoin files and non-blockstate paths would return without querying metadata. I also cached pack_format per resource pack instead of rechecking pack.mcmeta every blockstate lookup.

## Checklist

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinFileResourcePack.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinFileResourcePack.java
@@ -4,24 +4,26 @@ import static com.gtnewhorizon.gtnhlib.core.GTNHLibCore.MODEL_LOGGER;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.Reader;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
-import net.minecraft.client.Minecraft;
 import net.minecraft.client.resources.AbstractResourcePack;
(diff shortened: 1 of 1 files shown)


sunshinekitsune:
Rebased to resolve the conflict with 0648120b2bccd43696044c6ba3529210eacccafb.

Moving the checks to the top is a optimization that I originally did and it does help with performance but the way reading metadata is done still has the issues I described before, this PR is still relevant.

jss2a98aj (approved):
Stuff works as expected on my end with this. The changes also look fine to me.

PR #457, opened 2026-08-25
Add ChatComponentItemDisplayName

Eldrinn-Elantey:
## Summary

ChatComponentItemName wraps the name in square brackets, following vanilla func_151000_E, which is what chat wants but not what a GUI label wants. Outside of chat the brackets have to be stripped back off, and there is no way to ask for the bare name.

This adds a component that resolves the display name without brackets, plus a flag to keep only the contents of the last pair of parentheses ("Extruder Shape (Rod)" becomes "Rod"), which is a common shape for 1.7.10 item names when only the variant part is wanted.

Needed by GTNewHorizons/Applied-Energistics-2-Unofficial#1545, where an interface name suffix is built on a dedicated server and has to be localized on the client.

## Checklist
- [ ] I have tested this PR in DevEnv


Eldrinn-Elantey:
Closing this. The parentheses rule is GregTech-specific, so the component will be registered from GT5U instead, where it can also skip the brackets.

PR #464, opened 2026-08-30
Add inventoryCount to InventoryChangedEvent

Caedis:
## Summary
Allows mods like Amazing Trophies and BetterQuesting to use the value for detection instead of using the added event as a trigger to do their own scan.

image

## Checklist

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/event/InventoryChangedEvent.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/event/InventoryChangedEvent.java
@@ -4,8 +4,6 @@ import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.ItemStack;
 import net.minecraftforge.event.entity.player.PlayerEvent;
 
-import lombok.Getter;
-
 /**
  * Fired when a player's inventory changes (both client and server). Not cancelable. Subscribe to concrete
  * {@link ItemAdded} / {@link ItemRemoved} subclasses (1.7.10 Forge cannot register against abstract base without no-arg
--- a/src/main/java/com/gtnewhorizon/gtnhlib/event/InventoryChangedEvent.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/event/InventoryChangedEvent.java
@@ -1,9 +1,10 @@
 package com.gtnewhorizon.gtnhlib.event;
 
-import lombok.Getter;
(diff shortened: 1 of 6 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/event/InventoryChangedEvent.java
```diff
@@ -14,9 +16,14 @@ public abstract class InventoryChangedEvent extends PlayerEvent {
     /** The changed item (representative; identity ignores NBT). Shared across listeners - copy before mutating. */
     public final ItemStack item;
 
-    protected InventoryChangedEvent(EntityPlayer player, ItemStack item) {
+    /** Total amount of {@link #item} the player holds after this change (NBT-insensitive, same identity rules). */
+    @Getter
+    private final int inventoryCount;
```
Nikolay-Sitnikov: Is there a reason why `item` is `public final`, but this is `private final` with a `@Getter`?

(Personally, I don't like Lombok. Your opinion may differ.)

Caedis: It originally just was going to be a public final variable but i changed it to private with a getter.
I can do the same for `item` but its already being used in Amazing Trophies

We already use Lombok all over this repo

boubou19 (approved):

Jarnexis:
> ## Summary
> 
> Allows mods like Amazing Trophies and BetterQuesting to use the value for detection instead of using the added event as a trigger to do their own scan.
> image
> ## Checklist
> 
>     * [x]  I have tested this PR in DevEnv
> 
>     * [x]  I have tested this PR in Fullpack
> 
>     * [x]  This PR is in compliance with the [GTNH AI Policy](https://github.com/GTNewHorizons/GTNH-Dev-Doc/blob/master/AI_POLICY.md)
> 
>     * [ ]  This PR requires another PR in order to merge

Thanks for merging!

Just a quick question regarding future updates: I noticed that the chat messages in InventoryEventDebugHandler are hardcoded using ChatComponentText (e.g., "ADDED", "REMOVED", "(total "...).

Would you consider switching them to ChatComponentTranslation with proper localization keys in future versions so they can be translated into other languages? Or is this class strictly meant for internal debugging, meaning you'd prefer to keep the chat output and logs strictly in English?

PR #467, opened 2026-09-04
Add various debug logs to the model system

RecursivePineapple:
## Summary
The model system backport previously had very few logs in it, which made debugging broken models extremely difficult. This PR just adds some log messages when a GTNHLib config is enabled.

## Checklist

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinFileResourcePack.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinFileResourcePack.java
@@ -132,17 +132,19 @@ public abstract class MixinFileResourcePack extends AbstractResourcePack impleme
             }
         }
 
-        if (GTNHLibConfig.enableModelDebugLogs) {
-            if (this.nhlib$packFormat < PACK_FORMAT_MC_13_X) {
+        if (this.nhlib$packFormat < PACK_FORMAT_MC_13_X) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinFileResourcePack.java
```diff
@@ -131,6 +132,16 @@ public MixinFileResourcePack(File file) {
...
+        if (GTNHLibConfig.enableModelDebugLogs) {
+            if (this.nhlib$packFormat < PACK_FORMAT_MC_13_X) {
+                MODEL_LOGGER.warn(
+                        "Rejecting valid model at {} because its pack.mcmeta format is too old (is {}, must be >={})",
+                        resource,
+                        this.nhlib$packFormat,
+                        PACK_FORMAT_MC_13_X);
+            }
+        }
+
```
ah-OOG-ah, line 143: Can you merge this if? i.e. `if (too old) { if (logging) { log } return true } else return false`

ah-OOG-ah (changes requested):

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/models/MixinFileResourcePack.java
RecursivePineapple: Should be good now

ah-OOG-ah (approved):

PR #468, opened 2026-09-04
Add an interface that allows blocks to render multiple AABBs when hovered over

RecursivePineapple:
## Summary
This PR adds an interface that allows Blocks to render multiple preview AABBs when hovered over. This is purely cosmetic and doesn't change the hitscan logic at all.

## Checklist

### DIFF
--- a/src/main/java/com/gtnewhorizon/gtnhlib/GTNHLibConfig.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/GTNHLibConfig.java
@@ -28,6 +28,10 @@ public class GTNHLibConfig {
     @Config.RequiresMcRestart
     public static boolean blockSoundMixins;
 
+    @Config.Comment("Allow blocks to show multiple AABBs when hovered over")
+    @Config.DefaultBoolean(true)
+    public static boolean enableMultipleHoverAABBs;
+
     @Config.Comment("Enable item rendering modifications to allow rendering some items as translucent")
     @Config.DefaultBoolean(true)
     public static boolean enableTranslucentItemRenders;
--- a/src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
@@ -1,0 +1,54 @@
+package com.gtnewhorizon.gtnhlib.mixins.early;
+
+import net.minecraft.block.Block;
+import net.minecraft.client.multiplayer.WorldClient;
+import net.minecraft.client.renderer.RenderGlobal;
+import net.minecraft.util.AxisAlignedBB;
+import net.minecraft.util.MovingObjectPosition;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+
+import com.gtnewhorizon.gtnhlib.api.IBlockWithMultiplePreviewAABBs;
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
+import com.llamalad7.mixinextras.sugar.Local;
+
+@Mixin(RenderGlobal.class)
+public class MixinRenderGlobal_MultiplePreviewHitboxes {
+
+    @Shadow
+    public static void drawOutlinedBoundingBox(AxisAlignedBB p_147590_0_, int p_147590_1_) {
+        throw new UnsupportedOperationException("Implemented via mixin");
+    }
+
+    @Shadow
+    private WorldClient theWorld;
+
+    @WrapOperation(
+            method = "drawSelectionBox",
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lnet/minecraft/client/renderer/RenderGlobal;drawOutlinedBoundingBox(Lnet/minecraft/util/AxisAlignedBB;I)V"))
+    private void gtnhlib$drawMultiHitboxes(AxisAlignedBB p_147590_0_, int p_147590_1_, Operation<Void> original,
+            @Local(name = "block") Block block, @Local(argsOnly = true, name = "p_72731_2_") MovingObjectPosition hit,
+            @Local(name = "d0") double playerX, @Local(name = "d1") double playerY,
+            @Local(name = "d2") double playerZ) {
+        if (block instanceof IBlockWithMultiplePreviewAABBs aabbs) {
+            var bbs = aabbs.getPreviewAABBs(this.theWorld, hit.blockX, hit.blockY, hit.blockZ);
+
+            if (bbs != null) {
+                for (var bb : bbs) {
+                    drawOutlinedBoundingBox(
+                            bb.expand(0.002F, 0.002F, 0.002F).getOffsetBoundingBox(-playerX, -playerY, -playerZ),
+                            -1);
+                }
+
+                return;
+            }
+        }
+
+        original.call(p_147590_0_, p_147590_1_);
+    }
+}
--- a/src/main/java/com/gtnewhorizon/gtnhlib/api/IBlockWithMultiplePreviewAABBs.java
+++ b/src/main/java/com/gtnewhorizon/gtnhlib/api/IBlockWithMultiplePreviewAABBs.java
@@ -1,0 +1,17 @@
+package com.gtnewhorizon.gtnhlib.api;
+
+import java.util.List;
+
+import net.minecraft.util.AxisAlignedBB;
+import net.minecraft.world.World;
+
+import cpw.mods.fml.relauncher.Side;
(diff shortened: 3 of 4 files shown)


Nikolay-Sitnikov (changes requested):
This is already possible through Forge's `DrawBlockHighlightEvent` (see https://github.com/GTNewHorizons/BuildCraft/blob/master/src/main/java/buildcraft/core/render/BlockHighlightHandler.java for an example.)

Why do we need a mixin here?

RecursivePineapple:
@Nikolay-Sitnikov Code could call `RenderGlobal.drawSelectionBox` without going through the forge hook. WrapMethod generally doesn't have mixin conflicts, the only cost is the Object[] varargs alloc. There's not much of a difference, this is just more robust if mod code does something stupid.

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
```diff
@@ -0,0 +1,54 @@
...
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+
+import com.gtnewhorizon.gtnhlib.api.IBlockWithMultiplePreviewAABBs;
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
+import com.llamalad7.mixinextras.sugar.Local;
+
+@Mixin(RenderGlobal.class)
+public class MixinRenderGlobal_MultiplePreviewHitboxes {
```
Nikolay-Sitnikov, line 19: Why not just make this abstract, and make `drawOutlinedBoundingBox` `abstract`?

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
```diff
@@ -0,0 +1,54 @@
...
+            @Local(name = "d0") double playerX, @Local(name = "d1") double playerY,
+            @Local(name = "d2") double playerZ) {
+        if (block instanceof IBlockWithMultiplePreviewAABBs aabbs) {
+            var bbs = aabbs.getPreviewAABBs(this.theWorld, hit.blockX, hit.blockY, hit.blockZ);
+
+            if (bbs != null) {
+                for (var bb : bbs) {
+                    drawOutlinedBoundingBox(
+                            bb.expand(0.002F, 0.002F, 0.002F).getOffsetBoundingBox(-playerX, -playerY, -playerZ),
+                            -1);
```
Nikolay-Sitnikov, line 45: Why not `original.call(bb.expand(0.002F, 0.002F, 0.002F).getOffsetBoundingBox(-playerX, -playerY, -playerZ), p_147590_1_)`?

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
```diff
@@ -0,0 +1,54 @@
...
+                }
+
+                return;
+            }
+        }
+
+        original.call(p_147590_0_, p_147590_1_);
```
Nikolay-Sitnikov, line 52: This is a bit confusing because the "early return" branch is much larger than the "default" branch. This would look much better as an `else`, I'd say. (In general, only early return if the "early return" branch is short, and the main path is significantly longer.)

src/main/java/com/gtnewhorizon/gtnhlib/GTNHLibConfig.java
```diff
@@ -32,6 +32,10 @@ public class GTNHLibConfig {
     @Config.RequiresMcRestart
     public static boolean blockSoundMixins;
 
+    @Config.Comment("Allow blocks to show multiple AABBs when hovered over")
+    @Config.DefaultBoolean(true)
+    public static boolean enableMultipleHoverAABBs;
```
Nikolay-Sitnikov, line 37: Do we really want to allow people to disable this mixin, considering it is an API for other mods to rely on? (I guess it's not utterly gamebreaking, but it does seem unfortunate to have to force every mod that relies on this mixin to handle both the "mixin enabled" and "mixin disabled" case.)

Nikolay-Sitnikov (changes requested):

mitchej123:
As a note.  It's an Object[] allocation and boxing of every primitive.   Check in a profiler.  Depends on how hot it is if it matters.

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
RecursivePineapple: It's a static method, you can't make it abstract.

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
RecursivePineapple: This is just more precise. If you called the original then you'd have to go look at the target method to understand what this mixing does.

src/main/java/com/gtnewhorizon/gtnhlib/GTNHLibConfig.java
RecursivePineapple: It's standard for these optional mixins. I didn't make the sound mixins disableable and there was a mixin conflict with a third party mod. If someone disables the mixin then they just don't get the feature, same as the sound mixins. GTNHLib has to support more environments than normal.

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
RecursivePineapple: I prefer early returns/continues because the 'cognitive stack' you have to keep track of is smaller. Instead of having to look at the indent level you can just look at how far through the method/loop you are to figure out what assertions have been made.

RecursivePineapple:
This isn't a hot path, it's called at most once per frame. I'd be surprised if this shows up on a flamegraph at all.

mitchej123:
> This isn't a hot path, it's called at most once per frame. I'd be surprised if this shows up on a flamegraph at all.

lol once per frame can be hot, but if it doesn't show up on a profiler it's fine.  Just be mindful of it.

RecursivePineapple:
It should be fine, it only emits some gl_lines geometry for an aabb afaik

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
Nikolay-Sitnikov: If you're going to use `WrapOperation` to allow mixins to layer on top of each other, then you should maximize the amount that the mixins can layer on top of each other. `@WrapOperation` isn't just for suppressing mixin conflict errors, it's so mixins can actually layer on top of each other.

Also, I already have to go back to the original method to understand what the `0.002F` and `playerX/Y/Z` parameters mean.

src/main/java/com/gtnewhorizon/gtnhlib/GTNHLibConfig.java
Nikolay-Sitnikov: Alright, well at the very least this needs `@Config.RequiresMcRestart`.

src/main/java/com/gtnewhorizon/gtnhlib/mixins/early/MixinRenderGlobal_MultiplePreviewHitboxes.java
Nikolay-Sitnikov: This isn't an assertion, though. The "exceptional" early-return branch is 90% of the code, and it still suffers from the issue where I have a hard time telling what if-statement that return block belongs to. Early-returns help by literally not creating a new block of code, this allowing you to keep thinking of the code in a linear manner. This doesn't apply here because the "early-returns" path is longer than the entire function, and that return just makes me question "how many nested code blocks is that return breaking out of, again?"

The if-else at least avoids that second question - the control flow, although non-linear, can still be understood on one screen ("ah, both branches converge onto line XX, from where the flow is linear")

If you wish to keep using an early return, invert the condition, so the early-return path is shorter than the main path. (It's still not great, since you have to make a block for that early return, which makes the code more non-linear, but it's better that this code.)

Nikolay-Sitnikov:
@mitchej123 I once asked LlamaLad why `@WrapOperation` didn't use `MethodHandle`s. Apparently, all of these array allocations and boxing usually optimizes out.

PR #471, opened 2026-09-07
Capture system locale before mods overwrite it

Eldrinn-Elantey:
## Summary
SYSTEM_DEFAULT read the JVM locale during preInit. GregTech and ServerUtilities replace that locale with English there and never restore it, so number formatting ignored the locale the player actually runs.

GTNHLib now reads the locale during mod construction, before any mod pre-initialization runs.

## Checklist


YannickMG:
I have no context here, but are we sure that we want to write code to work around the locale being forced to english, rather than "fixing" those other mods so they do not alter the locale? Overriding the user locale sounds like an anti-pattern.

Eldrinn-Elantey:
Agreed, GregTech and ServerUtilities should restore the locale they replace, and I plan to open a PR against GregTech for it. Until a pack ships those fixes GTNHLib still reports the wrong locale, and reading it before mod loading costs two lines.

Eldrinn-Elantey:
https://github.com/GTNewHorizons/GT5-Unofficial/pull/7974
https://github.com/GTNewHorizons/ServerUtilities/pull/340

PR #473, opened 2026-09-16
Add WorldContextRegistry as API for mods like LittleBlocks

DarkShadow44:
## Summary

This PR adds `WorldContextRegistry`, a shared API for addressing virtual worlds created by mods such as [LittleBlocks](https://github.com/DarkShadow44/LittleBlocks) and [Metaworlds](https://github.com/Gordon-Frohman/Metaworlds-Mixins/).

These mods create child worlds that are rendered as part of a parent world, but many mods assume that every world can be identified by a Forge dimension ID alone. That assumption breaks for virtual worlds: Metaworlds subworlds share their parent's dimension ID, and different mods may use overlapping subworld IDs. Integrations such as ForgeMultipart support therefore currently require consumer-specific mixins.

The new registry gives each world a composite address consisting of:

- the host Forge dimension ID
- a namespace identifying the owning mod, such as `littleblocks`
- a subworld ID within that namespace

Ordinary worlds use subworld ID `0` and do not require a registered handler. Mods that own virtual worlds register a namespaced handler that can:

- derive an address from a world
- find the host world for a virtual world
- resolve a virtual world from its host and address
- enumerate the virtual worlds belonging to a host

This gives networking code a common way to identify and resolve the same virtual world on the server and client without knowing which mod owns it.

ForgeMultipart is the first integration being developed against this API. The related work is available here:

- ForgeMultipart: https://github.com/DarkShadow44/ForgeMultipart/tree/wip-worldregistry ([`c353c776`](https://github.com/DarkShadow44/ForgeMultipart/commit/c353c776c2e5087d96f4fa31498759d253c76cc5))
- LittleBlocks: https://github.com/DarkShadow44/LittleBlocks/tree/wip-worldregistry ([`d1baa0e4`](https://github.com/DarkShadow44/LittleBlocks/commit/d1baa0e4719c38cf32a22f1154efd48de1f22f3a))
- Metaworlds: https://github.com/DarkShadow44/Metaworlds-Mixins/tree/wip-worldregistry ([`f7106ee1`](https://github.com/DarkShadow44/Metaworlds-Mixins/commit/f7106ee1def31ce5c162e4fee55b1715459fae58))

Note that the above links are mostly AI generated, but they do serve as proof that the underlying API works as intended.

## Checklist


src/main/java/com/gtnewhorizon/gtnhlib/api/world/WorldContextRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+            if (this == other) return true;
+            if (!(other instanceof WorldAddress)) return false;
+            WorldAddress that = (WorldAddress) other;
+            return this.hostDimensionId == that.hostDimensionId && this.namespaceKey == that.namespaceKey
+                    && this.subId == that.subId;
+        }
+
+        @Override
+        public int hashCode() {
+            return (this.hostDimensionId * 31 + this.namespaceKey) * 31 + this.subId;
```
RecursivePineapple: This hash is pretty bad, could you use Fnv32 instead?

src/main/java/com/gtnewhorizon/gtnhlib/api/world/WorldContextRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+                        "Namespace '" + namespace + "' collides with '" + registered + "'; pick another");
+            }
+        }
+        namespaces.add(namespace);
+        handlers.add(handler);
+    }
+
+    /** {@link String#hashCode} is specified, so the key is stable across sides and independent of load order. */
+    private static int keyOf(String namespace) {
+        return namespace.hashCode();
```
RecursivePineapple: String hashes could theoretically conflict, is there a reason to not just send the string across the wire? If performance is a problem you could always intern the string and use reference equality.

src/main/java/com/gtnewhorizon/gtnhlib/api/world/WorldContextRegistry.java
DarkShadow44: Sure, done

src/main/java/com/gtnewhorizon/gtnhlib/api/world/WorldContextRegistry.java
DarkShadow44: Yeah I considered that, just an attempt at minimizing the size. Uses strings now

RecursivePineapple (approved):

